PR #6, opened 2017-09-02
Use Kerl instead of Curl

alaingilbert:
This fix the `GetUsedAddress` function.

Most of the logic come from the [JS library](https://github.com/iotaledger/iota.lib.js).


knarz:
Hey, thank you for this! 

Could you maybe try replacing as many of the numbers (6561, 243) with the respective constants? And I assume that some of the tests might have to be updated as well?

PR #7, opened 2017-09-07
use kerl for address generation and signing

Pomyk:
Most tests are updated but I have not checked them against other implementations.


knarz:
Thank you!

PR #12, opened 2017-10-25
Update curl implementations to use Curl-P-81

Pomyk:
except cl - this one is disabled for now.
Fix address checksum.


knarz:
LGTM, thank you

PR #14, opened 2017-11-08
Fix bundle hashing bugs from last MR.

Pomyk:
Tested with a successful giotan transfer.


Pomyk:
What's left to fix is input generation. After that I will add some more tests.

PR #17, opened 2017-12-01
mamutils for MAMs on the tangle

habpygo:
Made changes as was requested. Replaced stringutils folder with mamutils folder and simplified ToMAMTrytes function.


mamutils/sendreceivemam.go
```diff
@@ -0,0 +1,87 @@
...
+			//make it a space
+			number = 32
+		}
+		firstValue := number % 27
+		secondValue := (number - firstValue) / 27
+		trytesValue := string(TryteValues[firstValue]) + string(TryteValues[secondValue])
+		trytes = trytes + trytesValue
+	}
+
+	newTrytes := giota.Trytes(trytes)
```
knarz: Use `ToTrytes` to catch any errors and return them.

mamutils/sendreceivemam.go
```diff
@@ -0,0 +1,87 @@
...
+
+//ToMAMTrytes checks its validity and casts to giota.Trytes.
+func ToMAMTrytes(t string) (tr giota.Trytes) {
+
+	trytes := ""
+	TryteValues := giota.TryteAlphabet
+
+	for _, number := range t {
+		if number > 255 {
+			//make it a space
```
knarz: This should, at the very least, be mentioned in the description of the function. Silently converting input will just confuse people.

mamutils/sendreceivemam.go
```diff
@@ -0,0 +1,87 @@
...
+//FromMAMTrytes converts the MAM from giota.Trytes to a readable string
+func FromMAMTrytes(inputTrytes giota.Trytes) string {
+
+	outputString := ""
+	TryteValues := giota.TryteAlphabet
+
+	// Check if input is an even number of giota.Trytes
+	err := IsValidTrytes(inputTrytes)
+	if err != nil {
+		fmt.Println("Error: ", err)
```
knarz: If you encounter an error, don't just log it, but return it to the caller.

mamutils/sendreceivemam.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/iotaledger/giota"
+)
+
+//ToMAMTrytes checks its validity and casts to giota.Trytes.
```
knarz: The only thing this does is convert an ASCII (since everything above 255 is replaced by a space) string to trytes, so I'm not sure if this needs to be its own package since the same thing could be used for e.g. tags as well.

habpygo: No, indeed. I was wondering about this myself. So I will just use the file and take it out of the package.

mamutils/sendreceivemam.go
habpygo: Perhaps this could be documented in the README.md file?

mamutils/sendreceivemam.go
habpygo: Will do.

mamutils/sendreceivemam.go
habpygo: Will do.

habpygo: I will do the update later. Unfortunately they took the message board off-line, so I can only check it via the CLI, but that isn't good enough. Next week there will be another IOTA meet-up in Amsterdam, where I will ask to let the message board up for development.
Why aren't you coming to the meet-up? Last time they mentioned that it's the largest meetup in Europe.

mamutils/sendreceivemam.go
knarz: Add it to the description above the method and the readme if possible.

habpygo:
Waiting for new PR.

PR #20, opened 2017-12-09
Using TritHashLength constant in the library instead of hardcoded int value 243

0x13a:


0x13a:
is this project abandoned or is worth contributing?
cc @utamaro @knarz

knarz:
Thank you for the pull request.

0x13a:
Thanks for getting back

PR #21, opened 2017-12-11
[feature] added dependency management

0x13a:
Added golang/dep dependency management

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ You will need C compiler for linux to compile PoW routine in C.
 $ go get -u github.com/iotaledger/giota
 ```
 
-This library uses [godep](https://github.com/tools/godep) so you will need to install it first and then
+This library uses [golang/dep](https://github.com/golang/dep) so you will need to install it first and then
 
 ```
 $ dep ensure


knarz:
I appreciate the effort but seeing as this is supposed to be a library I'm against including the vendor directory in the repo.

0x13a:
@knarz having the vendor folder into the repository is commonly advised as good practice in the Golang ecosystem, however not strictly, here some pro/cons from https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory

**Pros**

- It's the only way to get truly reproducible builds, as it guards against upstream renames, deletes and commit history overwrites.
- You don't need an extra dep ensure step to sync vendor/ with Gopkg.lock after most operations, such as go get, cloning, getting latest, merging, etc.

**Cons**

- Your repo will be bigger, potentially a lot bigger, though prune can help minimize this problem.
- PR diffs will include changes for files under vendor/ when Gopkg.lock is modified, however files in vendor/ are hidden by default on Github.


What do you think?

knarz:
I am aware of the pros and cons and agree that for applications that best practice should be followed but I think that libraries should let their users decide.

0x13a:
I am not sure what is your point, but I won't hang more on this, I've removed the vendor folder. It's not a big deal at the end. Thanks

0x13a:
what do you think @knarz

0x13a:
should I expect a response anytime soon @knarz ?

PR #28, opened 2018-01-03
mamgoiota commit

habpygo:
@knarz We added tests and more functionality.


knarz:
Hey, thank you, that looks much better already but I still feel like this might as well be its own repo, just like the js MAM client is on its own as well.

habpygo:
OK, thanks. I will contact them to see how to get access to the iotaledger org. and add our own `mam.client.go` repo or something along those lines.

PR #31, opened 2018-01-07
added official dep tool lockfiles

deltaskelta:


deltaskelta:
@knarz what do you think about this?

PR #39, opened 2018-01-21
Fix bug in loop

abadojack:
`func BytesToTrits` currently results to a panic (`Index out of range`). Loop counter should be decremented instead of being incremented.


knarz:
Huh, thank you for catching this.

PR #40, opened 2018-01-24
Fixed bug in TestBundle

deltaskelta:
- ~~8ae05a4~~ 8d0ffc59bc3bb4cd841cadd30581cc4db9663ae1 introduced a subtle bug that caused TestBundle to hang, incTrits() logic
reworked to fix bug and avoid unnecessary logic.

- added check for testing.Short() flag to API tests to avoid unwanted network lag when
connecting to random public nodes.

- refactored TestBundle to be table driven.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ before_install:
 - gometalinter --install --update
 
 script:
-- travis_wait 30 go test -tags=avx -v -covermode=count -coverprofile=coverage.out -timeout 100m
+- travis_wait 30 go test -tags=avx -v -covermode=count -coverprofile=coverage.out -timeout 100m -short
 
 after_success:
 - gometalinter -e bindata --deadline=1000s ./...
--- a/api_test.go
+++ b/api_test.go
@@ -28,6 +28,10 @@ package giota
 import "testing"
 
 func TestAPIGetNodeInfo(t *testing.T) {
+	if testing.Short() {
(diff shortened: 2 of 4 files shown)


knarz:
What exactly was the problem introduced by that commit?

deltaskelta:
the break statement in `incTrits()` broke the `switch` and not the `for` loop. It took me forever to track down...

It caused the infinite conditionless `for` loop in `getValidHash` to never end 

but multiple conditions aren't needed since `else` an be completely avoided there with a single `if`

deltaskelta:
This slipped by because of the failing CI tests due to the API tests taking too long. As a temporary fix, I also added a check for the `-short` flag when testing. I think we should also add the `-short` flag to the travic-ci build file in order to test all non-API related functions with CI. 

I can add this to this PR if you agree

knarz:
Oh, I think you linked the wrong commit then, because that one does not touch the `incTrits()` function.

And sure, add that as well.

deltaskelta:
oh yeah wrong commit, edited the original comment. I'll add the `short` flag to CI and push a new commit shortly

CWarner818:
This fixed my issue on Darwin, thank you!

deltaskelta:
@CWarner818 no problem, with the `short` flag now on the CI tests we will be able to spot problems like this quickly in the future

PR #41, opened 2018-01-25
Fix#37 ignored error

deltaskelta:
- This PR builds on top of ba183f510af88785bb1ac3df631958f16d129530
- Fixed various call stacks to not panic and instead pass errors up the stack. 
- fixed ignored errors and passed them to the caller


luca-moser:
Already fixed in `next`. All omitted errors get returned to the caller.

PR #46, opened 2018-02-11
Add checkConsistency API call

cwarner818:
Adds support fo the checkConsistency API command which is used to determine whether to promote or reattach a transaction.

### DIFF
--- a/api.go
+++ b/api.go
@@ -197,6 +197,28 @@ func (api *API) GetNodeInfo() (*GetNodeInfoResponse, error) {
 	return resp, err
 }
 
+// CheckConsistency calls CheckConsistency API which returns true if confirming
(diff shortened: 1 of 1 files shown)


api.go
```diff
@@ -197,6 +197,28 @@ func (api *API) GetNodeInfo() (*GetNodeInfoResponse, error) {
...
+		Tails   []Trytes `json:"tails"`
+	}{
+		"checkConsistency",
+		tails,
+	}, resp)
+
+	return resp, err
+}
+
+// GetTrytesResponse is for GetTrytes API resonse.
```
knarz: Should be `CheckConsistencyResponse`.

cwarner818:
Fixed that, thank you.

knarz:
Thanks!

PR #47, opened 2018-02-11
Add new parameters to getTransactionsToApprove

cwarner818:
Add support for selecting a root transaction to start from when selecting tips as well as the number of random walks that should be performed.

### DIFF
--- a/transfer.go
+++ b/transfer.go
@@ -30,6 +30,9 @@ import (
 	"time"
 )
 
+// Number of random walks to perform. Currently IRI limits it to 5 to 27
+const NumberOfWalks = 5
+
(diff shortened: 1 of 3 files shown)


transfer.go
```diff
@@ -30,6 +30,9 @@ import (
 	"time"
 )
 
+// Number of random walks to perform. Currently IRI limits it to 5 to 27
+const NumberOfWalks = 5
```
knarz: Maybe rename this to `DefaultNumberOfWalks` if it's not a fixed value?

knarz:
Hey, thank you for this but I accepted https://github.com/iotaledger/giota/pull/53 instead, which has the same functionality as far as I can tell.

PR #53, opened 2018-02-13
Add Promote function

Pomyk:
additional params in GetTransactionsToApprove
fix typo in CheckConsistencyResponse


Pomyk:
Rebased on current master and changed the NumberOfWalks constant.

knarz:
Thank you!

cwarner818:
Shouldn't the promote function be a method on an `API` not a standalone function that requires an API be passed to it? I don't see why some functions are like that.

PR #56, opened 2018-02-17
Add conversion to and from bytes and Trytes

cwarner818:
This adds functions for converting from ASCII to Trytes and back. I would have loved to used the method name `ToTrytes` but that is already taken and would introduce different behavior if changed. 

This is based off the javascript implementation: https://github.com/iotaledger/iota.lib.js/blob/37c5a1b10d8d6d7c6d688443f7c5b474e64d7679/lib/utils/asciiToTrytes.js

### DIFF
--- a/trinary.go
+++ b/trinary.go
@@ -342,8 +342,8 @@ func ToTrytes(t string) (Trytes, error) {
 func ToString(t Trytes) string {
 	var output string
 	for i := 0; i < len(t); i += 2 {
-		v1 := t[i]
-		v2 := t[i+1]
+		v1 := strings.IndexRune(TryteAlphabet, rune(t[i]))
+		v2 := strings.IndexRune(TryteAlphabet, rune(t[i+1]))
 		decimal := v1 + v2*27
 		c := rune(decimal)
 		output += string(c)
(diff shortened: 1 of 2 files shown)


trinary.go
```diff
@@ -339,6 +339,32 @@ func ToTrytes(t string) (Trytes, error) {
...
+}
+
+func FromString(s string) Trytes {
+	var output string
+	chars := []rune(s)
+
+	for _, c := range chars {
+		v1 := c % 27
+		v2 := (c - v1) / 27
+		output += string(alphabet[v1]) + string(alphabet[v2])
```
Pomyk: This can panic on unicode characters. In my opinion it would be better to encode bytes not runes (but it's not how the js library works).

cwarner818: Agree about the panic, I will add a check and an error return value if it's supplied with non ASCII values. Aside from that I want to maintain compatibility with the JS library.

cwarner818:
I'm considering changing this PR to a generic byte <-> tryte conversion. Then if you wanted to encode a string you could call it with a cast to a `[]byte`. This would really make it more useful for other purposes.

cwarner818:
Any input on this? I would love to see it merged as this is a functionality that the javascript library has but we do not. It's very handy for sending JSON payloads in messages, for example.

luca-moser:
Please consider opening a new PR in case you miss functionality in the refactored library. `ASCIIToTrytes` and `TrytesToASCII` has been added. Thank you.

PR #60, opened 2018-02-24
Use google's SHA3 package

cwarner818:
This uses `github.com/golang/crypto/sha3` but modified for correct padding with IOTA's use of keccak. Only changes I made to the package were to move the SHA3 sub-directory to the root and deleted all the other packages. 

There are no benchmarks in giota so I started writing some. Currently using my spammer test program I noticed an approximate 10% improvement in spamming rates between the two keccak implementations.


luca-moser:
Google's sha3 package is used in the `next` branch using a forked package with the domain separation bits adjusted.

PR #62, opened 2018-02-25
Fixes PoW C not being able to be compiled on ARM

luca-moser:
This commit changes the unspecified signed/unsigned characteristics of the char values used inside the C transform and pow functions.

Through this commit, PoW C was able to be compiled on an ARMv8 Scaleway instance.

Signed-off-by: Luca Moser <moser.luca@gmail.com>


knarz:
Thank you!

PR #67, opened 2018-03-21
Let each PoW impl. benchmark the same tx and mwm

chenwei-tw:


knarz:
Maybe just make that tx a variable so it's a bit easier to change in the future?

maeck70:
I believe the different PoW types do test in different ways. Pow_Go and Pow_C only do a single test, whereas, Pow_SSE, Pow_C128 and others do multiple parallel tests.

chenwei-tw:
I think PoW needs more trytes for evaluation of each implementation, because different trytes give different performance.

For example, here is evaluation between SSE and AVX with 6000 trytes. (each sample measured elapsed time with 30 trytes, total 200 samples)


At least, I think a consistent tryte for each implementation is needed.

maeck70:
How are you suggesting we do that. Just cycle through an array of 30 Trytes of which the hash ends with '99999'?

luca-moser:
This changes are not compatible anymore with the refactored iota.go library. Please consider opening a new PR with your wanted changes. All tests in the refactored library use the same mwm.

PR #68, opened 2018-06-15
expose getvalidhash

veorq:
This exposes `Bundle.getValidHash()` as `Bundle.GetValidHash()`, for applications that use an external component for signing the bundle. Maybe not the optimal solution though,


knarz:
Thank you

PR #73, opened 2018-07-06
Added GetPowFunc() and GetPowFuncNames() and changed transactionTrinarySize from private to puplic 

muXxer:
Hi giota Team!

I needed a possibility to check from outside of the giota lib if a special Pow function exists.
Otherwise my build would fail on an ARM64 device, if I directly access the PoWSSE function in my code, which is needed on INTEL/AMD plattforms.

Another important information in our project was the transactionTrinarySize, which is the only constant that is private, so I changed it to puplic.


brunoamancio:
Merge this, please :)

shufps:
please!!!!

PR #87, opened 2018-10-17
Fix missing Signature/Message chunk for bundles with transaction coun…

HaBenN:
…t > 1


luca-moser:
This was fixed in the `next` branch.

PR #112, opened 2019-05-07
Use multithreading in MerkleCreate

muXxer:
# Description

Merkle tree creation takes a lot of time for big trees.
This PR speeds up the creation by using multithreading.

## Type of change

# How Has This Been Tested?

test_mam.go and test_merkle.go still pass.

# Checklist:

- [ ] I have written example code according to the contribution guidelines

### DIFF
--- a/merkle/merkle.go
+++ b/merkle/merkle.go
@@ -3,11 +3,13 @@ package merkle
 
 import (
 	"errors"
+	"runtime"
+	"sync"
 
 	. "github.com/iotaledger/iota.go/consts"
 	"github.com/iotaledger/iota.go/curl"
-	"github.com/iotaledger/iota.go/signing/legacy"
-	"github.com/iotaledger/iota.go/signing/utils"
+	signing "github.com/iotaledger/iota.go/signing/legacy"
+	sponge "github.com/iotaledger/iota.go/signing/utils"
 	. "github.com/iotaledger/iota.go/trinary"
 )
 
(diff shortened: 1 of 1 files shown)


merkle/merkle.go
```diff
@@ -106,35 +108,72 @@ func MerkleCreate(baseSize uint64, seed Trytes, offset uint64, security Security
...
-
-		key, err := signing.Key(subSeed, security, h)
-		if err != nil {
-			return nil, err
-		}
+	waitGroup := &sync.WaitGroup{}
+	errorChan := make(chan error)
+	waitChan := make(chan struct{})
+
+	numCPU := uint64(runtime.NumCPU())
```
luca-moser: Make the amount of goroutines to use configurable by the caller.

merkle/merkle.go
```diff
@@ -106,35 +108,72 @@ func MerkleCreate(baseSize uint64, seed Trytes, offset uint64, security Security
...
-		}
+	// Create signal when finished
+	go func() {
+		waitGroup.Wait()
+		waitChan <- struct{}{}
+	}()
 
-		treeIdx := MerkleNodeIndex(td, leafIndex, td)
-		copy(tree[treeIdx*HashTrinarySize:(treeIdx+1)*HashTrinarySize], address)
+	select {
```
luca-moser: This is actually schematically wrong. If any of the goroutine already has produced an error, then all other goroutines will be stuck forever trying to send into the error channel. Consider collecting the errors happening within the Merke tree generation and put them into a slice of errors (protected by a mutex of course). You can also drop the `waitChan` and just use the normal `WaitGroup` pattern. After you `wait()`ed, you can check if the error slice contains any errors.

PR #124, opened 2019-09-08
Added TransactionToTrits

muXxer:
# Description

I need the Trits of a transaction object in HORNET, and the existing TransactionToTrytes would convert the Trytes to Trits... and then to Trytes, which I would have to convert back to Trits.

## Type of change

# How Has This Been Tested?

Added a test case in "transaction_test.go"

# Checklist:

- [X] I have made corresponding changes to the documentation
- [X] I have written example code according to the contribution guidelines
- [X] I have added tests using ginkgo that prove my fix is effective or that my feature works


luca-moser:
Please add a new method instead of modifying the current one's signature, so that one still has the possibility to use TransactionsToTrytes.

PR #131, opened 2019-12-08
Optimize kerl

Wollac:
# Description

This PR improves the binary-ternary conversion used for Kerl, namely `KerlTritsToBytes` and `KerlBytesToTrits`.

- Previously, this was done with a trits<->bytes conversion, leading to uint64 multiplication/divisions for each of the 243 trits. This has been changed to a trytes<->bytes conversion that only needs to be performed for each of the 81 trytes. If necessary, these trytes can then be converted to trits using a LUT without any divisions.
- Introduce `KerlTrytesToBytes` and `KerlBytesToTrytes` when trytes are used directly.
- Adapt `AbsorbTrytes` and `SqueezeTrytes` to work on those functions directly.
- Optimize a few hot functions in `bigint` for speed.
- Change all usages of `Absorb` and `Squeeze` that included a trytes to trits conversion to the corresponding `AbsorbTrytes` and `SqueezeTrytes`.
- Add tests for new tryte-based functions.

On the tested machine this leads to a 2x - 3x speedup for all Kerl related functions. This has been tested with a few artificial benchmarks:
#### Before
```
BenchmarkKerlBytesToTrits-8    	  212587	      5089 ns/op
BenchmarkKerlTritsToBytes-8    	  521274	      2348 ns/op
BenchmarkKerl-8                	  139332	      8144 ns/op
BenchmarkKerlTrytes-8          	  117549	     10121 ns/op
BenchmarkGenerateAddress-8     	      97	  12075147 ns/oo
BenchmarkValidateSignature-8   	     194	   6050952 ns/op
```
#### After
```
BenchmarkKerlBytesToTrits-8    	  757398	      1366 ns/op
BenchmarkKerlTritsToBytes-8    	 1312630	       920 ns/op
BenchmarkKerl-8                	  328953	      3282 ns/o
BenchmarkKerlTrytes-8          	  324426	      3248 ns/op
BenchmarkGenerateAddress-8     	     238	   4918785 ns/op
BenchmarkValidateSignature-8   	     452	   2526877 ns/op
```
## Type of change

- [x] Enhancement (non-breaking change which adds functionality)

# How Has This Been Tested?

- [x] Tests in `iota.go` project test suite
- [x] Additional test cases for tryte based functions
- [x] Benchmarks to verify performance improvement

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
[...]

PR #132, opened 2019-12-10
Added platform specific bytesToTryteValues

muXxer:
# Description

bytesToTryteValues does a platform specific division now.
This is much faster on 32bit hardware.

## Type of change

- [x] Enhancement (non-breaking change which adds functionality)

# How Has This Been Tested?

Compiled on amd64 and arm

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have written example code according to the contribution guidelines
- [ ] I have added tests using ginkgo that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #134, opened 2019-12-16
Further kerl improvents

Wollac:
# Description

This PR adds further improvements to the binary-ternary conversion used for Kerl.

- Replace all unsafe operations. This prevents issues on big-endian hardware and is nicer go code :grinning:
- Add tests for edge cases; achieve 100% coverage in converter.
- Remove unused functions in `bigint`.
- Add `mipsle` to the architectures that require the 32bit version.
- Integrate @shufps idea, of using intermediate chunks of 6 trytes (the largest number of trytes that fits into a `uint32`). By doing bytes<->chunks and then chunks<->trytes conversion, `bigint` arithmetic is only needed in the first conversion. This reduces the number of bigint multiplication/divisions needed significantly compared to #131.

On the tested amd64 machine this leads to a 1.5x speedup for all Kerl related functions. This has been tested with a few artificial benchmarks:

#### Before
```
BenchmarkKerlBytesToTrits-8    	  757398	      1366 ns/op
BenchmarkKerlTritsToBytes-8    	 1312630	       920 ns/op
BenchmarkKerl-8                	  328953	      3282 ns/o
BenchmarkKerlTrytes-8          	  324426	      3248 ns/op
BenchmarkGenerateAddress-8     	     238	   4918785 ns/op
BenchmarkValidateSignature-8   	     452	   2526877 ns/op
````

#### After
```
BenchmarkKerlBytesToTrits-8      1706949               699 ns/op
BenchmarkKerlTritsToBytes-8      1972368               615 ns/op
BenchmarkKerl-8                   509046              2220 ns/op
BenchmarkKerlTrytes-8             467318              2191 ns/op
BenchmarkGenerateAddress-8           355           3369913 ns/op
BenchmarkValidateSignature-8         680           1721464 ns/op
```

## Type of change

- [x] Enhancement (non-breaking change which adds functionality)

# How Has This Been Tested?

- [x] Tests in `iota.go` project test suite
- [x] Additional tests for edge cases
- [x] Benchmarks to verify performance improvement

# Checklist:

- [x] My code follows the style guidelines of this project
- [ ] I have written example code according to the contribution guidelines


Wollac:
I am not a huge fan of the recent platform specific changes #132 as this adds parallel code to the library:
- The CI is currently only run on amd64, the 32-bit version is not tested.
- As I don't have the corresponding hardware, it would be could if someone could do some benchmarks whether the 32-bit version is still significantly faster than the 64-bit version on `386`, `arm` and `mips`. Otherwise that version could be removed.

Wollac:
@shufps did a few benchmarks for this:
- `tryteValuesToBytes` doesn't use any 64-bit operation, so there is no need for a special version.
- `arm`, `mips`, `mipsle` all use go runtime functions for the 64-bit divisions, so the special version will definitely be faster.
- `386` is faster without the special 32-bit version. Should we remove it from the `+build` flags, or keep it since it is the easiest to test in a CI environment.

shufps:
> * `386` is faster without the special 32-bit version. Should we remove it from the `+build` flags? Or keep it to be able to switch to `GOARCH=386` in the CI environment to run the tests?

I wouldn't care much about i386-version :) afaik Linux-support for i386 gradually is being removed from Linux and -Kernel^^

luca-moser:
> 386 is faster without the special 32-bit version. Should we remove it from the +build flags?
Lets move the `386` build flag to the 64bit version.

Wollac:
How can we run the tests with the 32-bit version?
Is it possible to switch the Travis architecture to `arm`? And then run the tests on `amd64` and `arm`?

Wollac:
That doesn't seem to be the case...
Travis CI can only test on amd64, ppc64le (IBM Power CPUs), s390x (IBM Z CPUs) and arm64 (ARMv8 compliant CPUs).

Any other ideas how this could be tested in the CI?
Can one somehow force that `bytes_to_trytes_32bit.go` is used although the build flags don't match?
Or we use `386` as a dummy (only for the CI) and run the tests twice on amd64 but once with `GOARCH=386`...

Wollac:
OK, the latest commit is what I am proposing:
- keep `386` build flag for the 32bit version
- use this to run CI tests on `amd64` architectures for the 32bit version
- on actual `386` this leads to slightly worse performance, but as @shufps mentioned, this should be more of a theoretical downside

kwek20:
WEN love for java </3

PR #155, opened 2020-04-17
Check if unused trits in tx value field are zero

muXxer:
# Description of change

Only the first 33 trits of the value field of a transaction are used in current IOTA implementation.
This PR adds a check, if the unused trits are set to zero.

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

Synced mainnet with HORNET

## Change checklist

- [X] My code follows the contribution guidelines for this project


Wollac:
I was wondering, would it maybe be more consistent to modify `TritsToInt` to report overflows?
This seems like a more general problem...

muXxer:
Sure we maybe could also do that, but the trailing trits shouldn't be used anyway, even if they would fit in an int64? :)

PR #162, opened 2020-05-03
Check value overflows in ValidBundle

muXxer:
# Description of change

This PR adds a check to ValidBundle to ensure that the balance mutations of a bundle would not overflow the total supply of IOTA.

## Type of change

## Change checklist

- [X] My code follows the contribution guidelines for this project

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -298,6 +298,10 @@ func ValidBundle(bundle Bundle) error {
 	for i := range bundle {
 		tx := &bundle[i]
 
+		if iotaGoMath.AbsInt64(tx.Value) > consts.TotalSupply {
+			return errors.Wrapf(ErrInvalidValue, "tx value (%d) overflows/underflows total supply", tx.Value)
+		}
+
 		totalSum += tx.Value
 		if iotaGoMath.AbsInt64(totalSum) > consts.TotalSupply {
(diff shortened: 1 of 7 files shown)


bundle/bundle.go
```diff
@@ -287,12 +291,22 @@ func ValidBundle(bundle Bundle) error {
...
 
 	sigs := make(map[Hash][]Trytes)
+	changes := map[trinary.Trytes]int64{}
 	k := kerl.NewKerl()
 
 	lastIndex := uint64(len(bundle) - 1)
 	for i := range bundle {
 		tx := &bundle[i]
+
 		totalSum += tx.Value
```
Wollac, line 301: Theoretically, this could also overflow, but for this `tx.Value` would need to be greater than the supply.
I am not 100% how the official validation rules are for transactions with value larger than the supply. So, either check tx.Value against the supply or check for overflows in the addition (maybe just using `big.Int` would be easiest in this case)

muXxer: Value is already checked during the parsing of the transaction. IMHO there is no need to check again?

Wollac: But maybe the the transaction was modified after, created via API/pure client code, ...
It is not checked in this function and if the validation rules forbid to have values > supply that is such an easy check to add making it much more robust without any drawback whatsoever

muXxer: done!

PR #163, opened 2020-05-26
Clean up signing and key derivation

Wollac:
# Description of change

- `signing/legacy` is only relevant for MAM and has therefore been moved inside the `mam` package.
- `merkle` is only relevant for MAM and has therefore been moved inside the `mam` package.
- Key derivation has been improved
  - Moved to own package `signing/key`
  - Former `signing.Key()` has been renamed to `key.Sponge()` and deprecated to better communicate that it only generates secure keys for true ternary sponge constructions (and not Kerl).
  - Introduced new `key.Shake()` to generate secure keys base on the SHAKE XOF.
  - All key derivation functions validate the correct form of the entropy and assure that the last trit is zero (for Kerl related hash functions).

## Type of change

- `key.Shake()` is tested against a hash from the Keccak reference implementation https://github.com/XKCP/XKCP for all security levels.
- `key.Sponge()` is tested against a key derived from the former implementation for all security levels and Kerl as well as Curl-P-81.
- Existing tests for address generation and signing all pass using `key.Sponge()`

Since CI checks are not reported for this PR, the corresponding test results can be found here:
https://travis-ci.org/github/iotaledger/iota.go/builds/692137743

## Change checklist


signing/key/key.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/iotaledger/iota.go/kerl/sha3"
+	sponge "github.com/iotaledger/iota.go/signing/utils"
+	. "github.com/iotaledger/iota.go/trinary"
+	"github.com/pkg/errors"
+)
+
+// Shake derives a private key of the given securityLevel from entropy using the SHAKE256 extendable-output function.
+// The entropy must be a slice of exactly 243 trits where the last trit is zero.
+// Shake derives its security assumptions from the properties of the underlying SHAKE function.
+func Shake(entropy Trits, securityLevel SecurityLevel) (Trits, error) {
```
semenov-vladyslav, line 47: I guess Shake256 can be turned into `sponge.SpongeFunction`, but for this task it would be overkill.

Wollac: Yes, you are right. It can be turned into a `sponge.SpongeFunction` by integrating the bin-ter conversion

PR #164, opened 2020-06-02
Merkle Tree support using SHAKE256 KDF

karimodm:
# Description of change

I've ported Hornet's Merkle code based on @Wollac's Shake KDF. This should become the new adopted standard when dealing with Merkle trees: at least until trinary is around. I've initially ported Hornet code, re-adapting a few bits to make sense in the context of iota.go.
In addition I've wrote a CLI tool to create and export Merkle tree using Hornet's marshaled format.

I am not so accustomed to write Go code, so bear with me ;)

## Type of change

- Trees created using the `CreateMerkleTree()` function have been tested to create proper structure as `*MerkleTree` struct.
- Each node of the Merkle Tree is a hash of children nodes using Kerl sponge.
- Leaves gets generated using a SHAKE256 KDF.

## Change checklist


tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
+package main
+
+import (
```
muXxer, line 3: I don't like the format of the whole file. the tabs/spaces are too short.

merkle/merkle_persistency.go
```diff
@@ -0,0 +1,157 @@
+package merkle
+
+import (
```
muXxer, line 3: When mention the original author of the code? :P

merkle/merkle_persistency.go
```diff
@@ -0,0 +1,157 @@
...
+			return err
+		}
+
+		mtl.Hashes = append(mtl.Hashes, trinary.MustBytesToTrytes(hashBuf, 81))
+	}
+
+	return nil
+}
+
+// storeMerkleTreeFile stores the merkle tree in a file.
```
muXxer, line 117: Uppercase

merkle/merkle_persistency.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	if err := fileBufWriter.Flush(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// loadMerkleTreeFile load the merkle tree from a file
```
muXxer, line 140: Uppercase and dot at the end

merkle/merkle_tree.go
```diff
@@ -0,0 +1,181 @@
...
+		input <- index
+	}
+
+	close(input)
+	wg.Wait()
+
+	return result
+}
+
+// Address generates an address deterministically, according to the given seed, milestone index and security level.
```
muXxer, line 148: "computeAddress generates..."

merkle/merkle_tree.go
```diff
@@ -0,0 +1,181 @@
...
+
+	address, err := trinary.TritsToTrytes(addressTrits)
+	if err != nil {
+		return "", err
+	}
+
+	return address, nil
+}
+
+// CreateMerkleTreeFile calculates an entire merkle tree
```
muXxer, line 180: Wrong function name, and dot at the end

tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
...
+    fmt.Println("'depth' cannot be lower than 1")
+    return
+  }
+
+  if securityLevel < 1 || securityLevel > 3 {
+    fmt.Println("'securityLevel' must be 1, 2 or 3")
+    return
+  }
+
+  if len(seed) != 81 {
```
muXxer, line 50: You should use the "guards" here.

tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
...
+  if err := merkle.StoreMerkleTreeFile(filePath, mt); err != nil {
+    return err
+  }
+
+  fmt.Printf("Merkle tree root: %v\n", mt.Root)
+
+  return nil
+}
+
+func main() {
```
luca-moser, line 26: Use the `log` package instead of using `fmt`.

tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+  "fmt"
+  "time"
+  "flag"
+
+  "github.com/iotaledger/iota.go/trinary"
+  "github.com/iotaledger/iota.go/merkle"
+)
+
```
luca-moser, line 11:
```suggestion
var (
  depth = flag.Int("depth", 0, "Depth of the Merkle tree to create")
  securityLevel = flag.Int("securityLevel", 0, "Security level of the private key used")
  seed = flag.String("seed", "", "Seed for leaves derivation")
  outputPath = flag.String("output", "", "Output file path")
)
```

tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
...
+  var depth int
+  var securityLevel int
+  var seed string
+  var outputPath string
+
+  flag.IntVar(&depth, "depth", 0, "Depth of the Merkle tree to create")
+  flag.IntVar(&securityLevel, "securityLevel", 0, "Security level of the private key used")
+  flag.StringVar(&seed, "seed", "", "Seed for leaves derivation")
+  flag.StringVar(&outputPath, "output", "", "Output file path")
```
luca-moser, line 36:
```suggestion
```

tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
...
+
+  flag.Parse()
+
+  if depth < 1 {
```
luca-moser, line 40:
```suggestion
  if *depth < 1 {
```

tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
...
+    fmt.Println("'depth' cannot be lower than 1")
+    return
+  }
+
+  if securityLevel < 1 || securityLevel > 3 {
```
luca-moser, line 45:
```suggestion
  if *securityLevel < 1 || *securityLevel > 3 {
```

tools/merkle_calculator/merkle_tree_calculator.go
luca-moser, line 50:
```suggestion
  if len(*seed) != 81 {
```

tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
...
+    fmt.Println("'securityLevel' must be 1, 2 or 3")
+    return
+  }
+
+  if len(seed) != 81 {
+    fmt.Println("'seed' must be a string of 81 trytes")
+    return
+  }
+
+  if outputPath == "" {
```
luca-moser, line 55:
```suggestion
  if *outputPath == "" {
```

tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
...
+    fmt.Println("'output' is required")
+    return
+  }
+
+  fmt.Printf("calculating %d addresses...\n", 1 << uint(depth))
```
luca-moser, line 60:
```suggestion
  fmt.Printf("calculating %d addresses...\n", 1 << uint(*depth))
```

tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
...
+
+  ts := time.Now()
+
+  createMerkleTreeFile(outputPath, trinary.Hash(seed), securityLevel, depth)
```
luca-moser, line 64:
```suggestion
  createMerkleTreeFile(*outputPath, *seed, *securityLevel, *depth)
```

luca-moser: On a side note, I'd actually simply move the content of the function into `main()`.

merkle/merkle_persistency.go
karimodm: Sorry for that!

tools/merkle_calculator/merkle_tree_calculator.go
karimodm: Not sure what you mean here...

merkle/merkle_tree.go
```diff
@@ -0,0 +1,196 @@
+/*
+Package merkle provides functions and types to deal with the creation and storage of
+Merkle trees, using the secure SHAKE256 KDF implemented in the signing/key package:
+thus not being affected by the the infamous M-Bug.
+
+The functions exported by the package are:
```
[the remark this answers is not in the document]
karimodm: Done ;)

merkle/merkle_tree.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/iotaledger/iota.go/consts"
+	"github.com/iotaledger/iota.go/kerl"
+	"github.com/iotaledger/iota.go/signing"
+	"github.com/iotaledger/iota.go/signing/key"
+	"github.com/iotaledger/iota.go/trinary"
+)
+
+// MerkleTree contains the merkle tree used for the coordinator signatures.
+type MerkleTree struct {
+	// depth of the merkle tree
```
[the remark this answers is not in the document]
karimodm: Fixed also in other struct members and comments.

tools/merkle_calculator/merkle_tree_calculator.go
karimodm: These vars are not pointers. Why do you attempt to dereference?

tools/merkle_calculator/merkle_tree_calculator.go
karimodm: See above

tools/merkle_calculator/merkle_tree_calculator.go
karimodm: Not a pointer

tools/merkle_calculator/merkle_tree_calculator.go
karimodm: Not a pointer

tools/merkle_calculator/merkle_tree_calculator.go
karimodm: Included function code into main

tools/merkle_calculator/merkle_tree_calculator.go
luca-moser: Look at suggestion which moves the flags to a `var ()` block

karimodm: Why is it better?

tools/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,67 @@
+package main
```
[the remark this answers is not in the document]
karimodm: `go fmt`, neat! Didn't know!

karimodm:
All your comments should be fixed: except for @luca-moser 's `flag.Int` vs my `flag.IntVar`.
I just want to know why the former is better.

luca-moser:
Because then the program is shorter and the flags are on top of the file which makes it easier to read.

tools/merkle_calculator/merkle_tree_calculator.go
karimodm: Oh I see: guards.IsTrytesOfExactLength ;)

tool/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,61 @@
...
+		log.Panicf("'depth' cannot be lower than 1")
+		return
+	}
+
+	if *securityLevel < 1 || *securityLevel > 3 {
+		log.Panicf("'securityLevel' must be 1, 2 or 3")
+		return
+	}
+
+	if !guards.IsTrytesOfExactLength(*seed, consts.HashTrytesSize) {
```
muXxer, line 36: You can use `guards.IsTransactionHash` instead

merkle/merkle_tree.go
```diff
@@ -0,0 +1,195 @@
...
+				}
+				resultLock.Lock()
+				result[int(index)] = address
+				resultLock.Unlock()
+			}
+		}()
+	}
+
+	for index := 0; index < count; index++ {
+		input <- MilestoneIndex(index)
```
muXxer, line 74: You removed my code to calculate the progress and especially the remaining time.
That will be a lot of fun for the user to wait several hours without feedback. :)

Can you please add a consumer function for the calculated progress? So the user can decide if he wants to print it or not.

tool/merkle.go
```diff
@@ -0,0 +1,60 @@
+/*
+Ported from https://github.com/gohornet/hornet codebase.
+Original authors: muXxer <mux3r@web.de>
+                  Alexander Sporn <github@alexsporn.de>
+                  Thoralf-M <46689931+Thoralf-M@users.noreply.github.com>
+
+
+Package tool provides miscellneous functionality to deal with IOTA types.
```
muXxer, line 8: miscellaneous

tool/merkle.go
```diff
@@ -0,0 +1,60 @@
+/*
+Ported from https://github.com/gohornet/hornet codebase.
```
muXxer, line 2: I'm not sure if these two functions should be part of the tool package, or the merkle package....

IMHO it fits more in the merkle package, but maybe @luca-moser should decide that.

tool/merkle_calculator/merkle_tree_calculator.go
```diff
@@ -0,0 +1,61 @@
...
+	if *outputPath == "" {
+		log.Panicf("'output' is required")
+		return
+	}
+
+	log.Printf("calculating %d addresses...\n", 1<<uint(*depth))
+
+	ts := time.Now()
+
+	mt := merkle.CreateMerkleTree(trinary.Hash(*seed), *securityLevel, *depth)
```
luca-moser, line 50:
```suggestion
	mt := merkle.CreateMerkleTree(*seed, *securityLevel, *depth)
```

trinary.Hash is already a string

tool/merkle.go
```diff
@@ -0,0 +1,60 @@
+/*
```
luca-moser, line 1: Please adjust this header: The description of the package shouldn't use `/* */` and instead just `//`.

Make a new line between `Package tool provides miscellneous functionality to deal with IOTA types.` and the above.

Check this out as a reference: https://github.com/golang/go/blob/master/src/compress/gzip/gunzip.go

tool/merkle.go
luca-moser: I agree, this two functions should reside in the merkle package.

Wollac: Actually, it was me who made the change request to move them outside the merkle package :stuck_out_tongue: 

We have `merkleTree.WriteTo` which takes an `io.Writer` and this is the key library function offering standard go support to use it together with files, mem, network...
This function is just a trivial wrapper around `WriteTo` with a file as a writer. Also, there are use-case specific options for this (like filemode, buffered vs non-buffered, compression or not, keeping file open...), which cannot be changed. That's why I don't see this is library functionality.

But its absolutely @luca-moser's call, I just wanted to explain my motivation.

luca-moser: You're right. The reason why I wanted this to be moved is that I find it odd to have `tools.LoadMerkleTreeFile` because `tools` is a very generic package name. @karimodm you can leave the two functions in place.

Wollac: That is actually a good point. The Merkle stuff is actually the first and only "tool" we have beside the iota.go lib. So at some point we should think about how to deal with this. Do we want to have this in a new repo, in a new folder, in a new package...

merkle/merkle_tree.go
karimodm: Yeah, removed your code as it was turned into a lib function.
I've addressed your comment by turning`CreateMerkleTree` into a variadic function that may receive a slice of progressCallbacks (only the first one will be used though).

tool/merkle.go
karimodm: Why this was appoved then? https://github.com/iotaledger/iota.go/pull/163/files#diff-2399c9a33af208ffcfaf1b9c867265a3

merkle/merkle_tree.go
```diff
@@ -0,0 +1,196 @@
...
+
+// MerkleTreeLayer contains the nodes of a layer of a merkle tree.
+type MerkleTreeLayer struct {
+	// level of the layer in the tree
+	Level int
+	// nodes of the layer
+	Hashes []trinary.Hash
+}
+
+type MilestoneIndex uint32
```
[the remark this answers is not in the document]
karimodm: I think for the same reason why Trytes is a type instead of simple string. It better determines what object we are dealing with.
You are right about it not being used outside of this packge: I've unexported it.

merkle/merkle_persistency.go
```diff
@@ -0,0 +1,157 @@
...
+		mtl := &MerkleTreeLayer{}
+
+		if err := mtl.Unmarshal(buf); err != nil {
+			return err
+		}
+
+		mt.Layers[mtl.Level] = mtl
+	}
+
+	mt.Root = mt.Layers[0].Hashes[0]
```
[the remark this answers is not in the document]
karimodm: Accounting for all read problems is pretty annoying in this context: you might not be able to read all the hashes needed for a specific layer for instance; the `MerkleTreeLayer.ReadFrom` function would although need to know the depth of the layer that is trying to load to know if all hashes have been unmarshaled: hashesCount is also retrieved from the Reader.
To keep it sane I've introduced just these two checks:

`mt.depth == 0`
and
`lengthLayers != 1<<mt.Depth`

merkle/merkle_tree.go
```diff
@@ -0,0 +1,196 @@
...
+// Address generates an address deterministically, according to the given seed, milestone index and security level.
+// a modified key derivation function is used to avoid the M-bug.
+func computeAddress(seed trinary.Hash, index MilestoneIndex, securityLvl int) (trinary.Hash, error) {
+
+	subSeedTrits, err := signing.Subseed(seed, uint64(index), kerl.NewKerl())
+	if err != nil {
+		return "", err
+	}
+
+	keyTrits, err := key.Shake(subSeedTrits, consts.SecurityLevel(securityLvl))
```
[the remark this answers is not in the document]
karimodm: I don't think we should encourage the creation of a MerkleTree with different KDFs if we don't have a suitable alternative to SHAKE.

merkle/merkle_tree.go
```diff
@@ -0,0 +1,196 @@
...
+
+	last := addresses
+	layers = append(layers, last)
+
+	for i := depth - 1; i >= 0; i-- {
+		last = calculateNextLayer(last)
+		layers = append(layers, last)
+	}
+
+	// reverse the result
```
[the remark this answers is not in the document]
karimodm: I've simplified the function quite a lot thanks to your suggestion.

merkle/merkle_tree.go
```diff
@@ -0,0 +1,196 @@
...
+
+type MilestoneIndex uint32
+
+// calculateAllAddresses calculates all addresses that are used for the merkle tree of the coordinator.
+func calculateAllAddresses(seed trinary.Hash, securityLvl int, count int) []trinary.Hash {
+	resultLock := &sync.Mutex{}
+	result := make([]trinary.Hash, count)
+
+	wg := sync.WaitGroup{}
+	wg.Add(runtime.NumCPU())
```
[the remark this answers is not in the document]
karimodm: I've actually realized that `iota.go` already uses `runtime.NumCPU` in `pow_go.go`. Should we keep it?
Regarding allowing the user to choose the level of parallelism of the merkle creation tool I think it is overkill: not for the tool itself, but for the library to account for such a user decision; it would be true for PoW too at the moment, a way more frequent operation.

merkle/merkle_tree.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+// MerkleTree contains the merkle tree used for the coordinator signatures.
+type MerkleTree struct {
+	// depth of the merkle tree
+	Depth int
+	// root address of the merkle tree
+	Root trinary.Hash
+	// merkle tree layers indexed by their level
+	Layers map[int]*MerkleTreeLayer
```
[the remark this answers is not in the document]
karimodm: Yep, makes sense. There is just a gotcha in the code: Layers will be actually sized depth + 1 (the Root layer is not accounted for).

merkle/merkle_tree.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// calculateNextLayer calculates a single layer of the merkle tree used for coordinator signatures.
+func calculateNextLayer(lastLayer []trinary.Hash) []trinary.Hash {
+
+	resultLock := &sync.Mutex{}
+	result := make([]trinary.Hash, len(lastLayer)/2)
+
+	wg := sync.WaitGroup{}
+	wg.Add(runtime.NumCPU())
```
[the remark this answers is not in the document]
karimodm: See comment above regarding `GOMAXPROCS`

merkle/merkle_tree.go
```diff
@@ -0,0 +1,196 @@
...
+	wg.Wait()
+
+	return result
+}
+
+// Address generates an address deterministically, according to the given seed, milestone index and security level.
+// a modified key derivation function is used to avoid the M-bug.
+func computeAddress(seed trinary.Hash, index MilestoneIndex, securityLvl int) (trinary.Hash, error) {
+
+	subSeedTrits, err := signing.Subseed(seed, uint64(index), kerl.NewKerl())
```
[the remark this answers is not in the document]
karimodm: I didn't like `computeAddress` to depend on the `signing` functions to perform a reset of the sponge after they are done.

merkle/merkle_tree.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	return address, nil
+}
+
+// CreateMerkleTreeFile calculates an entire merkle tree
+func CreateMerkleTree(seed trinary.Hash, securityLvl int, depth int) *MerkleTree {
+
+	addresses := calculateAllAddresses(seed, securityLvl, 1 << uint(depth))
+	layers := calculateAllLayers(addresses)
```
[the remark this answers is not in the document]
karimodm: Changed the logarithm for what you suggested: `bits.Len(uint(len(addresses))) - 1`

merkle/merkle_tree.go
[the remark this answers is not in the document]
Wollac: The only realistic use-case I could see for this is, if someone needs to regenerate the Merkle tree for the current Compass. To maybe recover a lost file or to extend an old one....
But I have no idea, If you don't think this is realistic let's force Shake

merkle/merkle_tree.go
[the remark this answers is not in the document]
Wollac: OK, then I would say let's make it `Layers []*MerkleTree` allocated to the correct size. This should even make the code a bit more readable.

merkle/merkle_tree.go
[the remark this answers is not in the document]
Wollac: I was actually proposing to make it configurable, because the PoW has an optional parameter like this.
I really don't think that `runtime.NumCPU()` is a good idea, because it always only returns the number of logical CPUs independently of what could actually be used. `GOMAXPROCS` is more flexible here, and I guess the user can also override it with an environmental variable. Which might indeed render it unnecessary to have this a parameter :thinking: 
@luca-moser what do you think?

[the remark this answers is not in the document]
luca-moser: I think as with the PoW funcs, we should add an optional `paralleism ...int` parameter in case one wants to use a specific amount of concurrent goroutines for the Merkle tree generation. Per default it should use `runtime.GOMAXPROCS(0)` and not `runtime.NumCPU()` since `GOMAXPROCS` could have been changed and be different from `runtime.NumCPU()` (its default value however is the same as `runtime.NumCPU()`).

merkle/merkle_tree.go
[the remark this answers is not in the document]
muXxer: Let's enforce Shake.
For the old merkle tree, compass can be used.

luca-moser:
Wonderful. Thank you Andrea for the PR.

PR #180, opened 2020-09-08
Curl improvements

Wollac:
:warning: This PR removes all [mam/v1](https://github.com/iotaledger/iota.go/tree/master/mam/v1) related code. :warning: 

MAM uses undocumented Curl calls and it several contains bugs that would be hard to fix.

# Description of change

This PR improves the Curl performance by optimizing the native go implementation as well as providing amd64 assembly. It removes the cgo dependency.  

On amd64 (assembler) compared against the previous cgo implementation:
```
CurlTransaction-48  2.62ms ± 4%  1.55ms ± 3%  -40.86%
CurlHash-48          152µs ± 2%    49µs ± 2%  -68.00%
```

On 386 (go) against the previous cgo implementation:
```
CurlTransaction-48  2.94ms ± 3%  2.72ms ± 3%   -7.66%
CurlHash-48          172µs ± 2%    85µs ± 2%  -50.44%
```

It also integrates https://github.com/iotaledger/hive.go/blob/master/batchhasher/bct_curl.go with additional optimizations.

On amd64 (assembler) compared against the hive.go implementation:
```
CurlTransaction-48  10.7ms ± 5%  6.13ms ± 5%  -42.71%
```

## Type of change

The existing ginkgo tests have been extended to test for more cases.

## Change checklist


curl/bct/transform_amd64.s
```diff
@@ -0,0 +1,76 @@
+// +build amd64,!gccgo,!appengine
```
Wollac, line 1: The `bct.Curl` assembler version provides a speed-up of about 10%. Does this speed-up justify the introduction of assembler code which is much harder to review and maintain?

Wollac:
I am referring to the [mam/v1](https://github.com/iotaledger/iota.go/tree/master/mam/v1) package in this repository. So I was naturally assuming this it was MAMv1. But @kuhlmannmarkus, you are right, it might be MAMv0.

curl/bct/transform_amd64.s
luca-moser: Since you've already also introduced assembler code for Curl I don't see any reason why it then should be left out for bct.Curl. (Maintenance complexity stays the same)

luca-moser (approved):
Looks good to me.

Issue #198, opened 2020-10-08
Make Curl single hasher apply the curl function in a batched manner.


c-mnd:
## Description

A batch-like workflow can be used in the singe hasher as well.

## Motivation

It reduces energy consumption of Curl single hasher.

## Requirements

1. replace single hashing algo/implementation by a new one
2. add transform functions to support avx, avx2

## Are you planning to do it yourself in a pull request?

Yes.

PR #199, opened 2020-10-08
Curl sb (single-batched)

c-mnd:
# Description of change

Closes issue #198 .
Reduce energy consumption of Curl single hasher, by applying the curl function in a batched manner.
If you know curl hashing, it's quite obvious for the first round: you can rotate the state by 364, then use opposite trits as inputs for the curl function. The resulting state will have resorted indices, but their offsets are consistent throughout the new state. 
Rotate again to have index 364 being in opposition to index 0, then again 'curl' them. Repeat, until rounds are done.
Finally, sort the state.
Similar to bct, it encodes one trit as two bits on different uint64 arrays,
n (Negative) and p (Positive): 0 -> n=0, p=0; 1 -> n=0, p=1; -1 -> n=1, p=0.
Top level arrays store 243 trits each, 2nd level arrays  get filled up 3*64+51.
This way, when doing 81 rounds, resorting can be done by and-mask operations.

## Type of change

Besides additional ginkgo tests, I'm running a HORNET node with the change successfully so far.

## Change checklist


Wollac:
This idea is really impressive!

I've re-implemented your proposal here https://github.com/Wollac/iota.go/tree/curl-sb 
It tries to address the most relevant issues I see with this PR: I believe, it improves code readability by being more "idiomatic go". This approach is really tailored for 81-round Curl, so it removes the deprecated Curl-P-27 completely. It does not rely on C-code to improve maintainability and portability.

By doing so, https://github.com/Wollac/iota.go/tree/curl-sb still achieves a 4x-5x speedup on amd64 over current `master`: 
```
CurlTransaction     1.53ms ± 3%  0.32ms ± 2%  -79.27%  (p=0.000 n=12+11)
CurlHash            48.2µs ± 3%  11.9µs ± 5%  -75.27%  (p=0.000 n=12+12)
```

Of course, AVX2 Cgo code in #199 are still a bit faster. (I've measured 10%-15% on my machine.) But a cleaner native go implementation might justify this small drawback.

c-mnd:
Thank You. I knew, the algorithm is ok, but my implementation was of course lousy.
I'll read your rewrite and learn!

c-mnd:
Yeah, this way it's a lot cleaner.
Compiler hints like these would have never come to my mind, great.
What I don't get: calling it a _left_ rotation, while the 'leading' shift is `shrInto`. Is it intentionally (looking at it as a higher level array context)?

Wollac:
Yeah, what I mean is that when calling `rotateState` with `1`, it will move the trit at index i to index i-1. So, the state elements are rotated to the left, which is exactly what you want when comparing it to the array based `transform`.
However, you are absolutely right, on a bit level it is doing a `ROR`.

c-mnd:
In an earlier implementation I called it downshift (`shd`) when the movement was towards lower indices, upshift (`shu`) for the opposite. This way it should work for both LTR (tits, trytes, enumerations) and conventional RTL numbers, lowlevel as well as higher.
Still, that introduced new terms and the necessity for imaginary translation even grew.
Finally, I just dropped naming the rotations direction, as these are not hard-to-understand code areas.

So the information you might want to provide in the comments is that there's _motion towards lower indices_.

c-mnd:
👌

c-mnd:
When possible and semantics don't suffer, operations on state members should be done in the order of their definition, for better reading and maintaining.
Hopefully I caught them all: https://github.com/c-mnd/iota.go/commit/42363bf8b26917942bfe32a24540473c2317fc59.

Wollac:
Yes, I fully agree! This was just an oversight on my side as I ordered `p` before `n`, but apparently forgot a few lines

luca-moser:
Closed via #218

PR #201, opened 2020-10-09
Add draft RFC-20 Bech32 address format

Wollac:
# Description of change

- Port Bech32 address format from https://github.com/Wollac/iota-crypto-demo
- Introduce `Address` interface containing a `Bech32` method.
- Add RFC examples as test cases.


luca-moser (approved):
lgtm

PR #208, opened 2020-10-31
Add draft RFC-43 message PoW

Wollac:
# Description of change

- Port Curl-based PoW for arbitrary binary data from https://github.com/Wollac/iota-crypto-demo
- Remove no longer working legacy Cgo `pow` code
- Add simple Go miner computing about 250 kH/s per thread on recent amd64 hardware


luca-moser (approved):
lgtm

PR #218, opened 2020-11-16
Curl improvements

### LINKED ISSUE #198: Make Curl single hasher apply the curl function in a batched manner.

Wollac:
# Description of change

The PR is a re-implementation of @c-mnd's #199. It switches the Curl state from a trit (`int8`) array to two bit arrays allowing for the Curl S-box to be applied to 64 trits in one call. 

In the current `master`, one Curl round performs the following transformation:
```
for i ← 1 to 729
   x ← S[1]
   S ← rot(S)
   y ← S[1]
   N[i] ← g(x,y)
 S ← N
```
Each element of the state `S` is combined with its rotated counterpart using the S-box `g`. 
The PR replaces this with the equivalent approach of rotating just once and applying the S-box on the entire state:
```
N ← rot(S)
S ← g(S,N)
```
The only difference then is, that the trits are at the wrong position. Successive trits are now an opposite rotation apart. This rotation offset adds up over the rounds and needs to be reverted in the end. In order to able to reorder the trits efficiently using only bit-mask operations, the Curl state is stored as three chunks of 243 trits each.

This achieves a 4x-5x speedup on amd64 over current `master`: 
```
CurlTransaction     1.53ms ± 3%  0.31ms ± 1%  -79.69%
CurlHash            48.2µs ± 3%  11.6µs ± 1%  -75.99%
```

Since this version is optimized for Curl-P-81, this PR also removes Curl-P-27 support completely. Curl-P-27 was no longer used in the protocol for quite some time.

This PR closes #198 

### `curl/bct`

This PR also contains some miner cleanups of `curl/bct`:
- The returned type is now exposed
- The `in`/`out` methods have been changed to include a small performance increase:
```
CurlTransaction     6.19ms ± 2%  5.94ms ± 2%  -3.95%
CurlHash             317µs ± 2%   296µs ± 2%  -6.78%
```

## Type of change

- Enhancement
- Breaking change (removal of the previously deprecated Curl-P-27)

- A test set of 100 vectors each of several Curl-P-81 use-cases, has been created with the current reference implementation. The validation of these hashes has been included in the regular test suite.
- Added tests for other situations to the Curl and BCT-Curl test suite.

## Change checklist


luca-moser (approved):
Looks good to me. I'll try to merge this code into the actual phase 2 branch.

PR #238, opened 2021-01-16
Avoid using Serializable interface as the map key for the dust allowance calculations.

alexsporn:
The Ed25519Address used internally is a pointer and it will fall back to pointer equality instead of the value if used as a map key. Instead use the String() method of each address.
This also changes the DustAllowanceFunc to use the Address interface instead of the more generic Serializable since we are always working with addresses here.


luca-moser (approved):
thanks, lgtm

PR #245, opened 2021-01-20
Fix Curl.CopyState

Wollac:
# Description of change

Fixes an issue with the `Curl.CopyState` returning the wrong state. As this function is used in the PoW mining, it leads to invalid nonces...

## Type of change


luca-moser:
Thanks

PR #248, opened 2021-01-25
Add RFC-28 (ZIP-215) compatible Ed25519

Wollac:
# Description of change

- Provide a drop-in replacement for `crypto/ed25519` that is compatible with [RFC-28](https://github.com/iotaledger/protocol-rfcs/pull/28).
- It uses `filippo.io/edwards25519` for the edwards25519 elliptic curve implementation. This package "should be a safer, faster, and more powerful alternative" to using a fork of `crypto/ed25519/internal/edwards25519`.


luca-moser:
lgtm, thanks!

PR #252, opened 2021-01-28
Create a v2 module for Chrysalis

Wollac:
# Description of change

- This essentially introduces `iota.go/v2` as a mayor (breaking) version. But it allows packages to use (directly or indirectly) legacy as well as Chrysalis.
- It removes all the `legacy` packages.

With this `iota.go/v2` requires `iota.go` (v1) for Curl and all the ternary stuff. Probably, this should be cleaned up, so that `iota.go/v2` is self-contained.


luca-moser (approved):
Nice deletions 👍

PR #269, opened 2021-03-10
Do not use indices LUT in Batched Curl

Wollac:
# Description of change

This PR gets rid of the LUT `curl.Indices` in the batched Curl transpose and computes it value on the fly. In order to avoid costly modulo operations or conditionals it unrolls an even number of iterations in the state loop. That way, the indices can be computed by alternately add 364 or subtract 365.

The leads to a 10% speedup on `amd64` compared to the current master:
```
CurlTransaction     3.18ms ±10%  2.86ms ± 5%  -10.29%
CurlHash             115µs ± 4%   106µs ± 6%   -7.85%
```
By default, `amd64` uses the Assembly version; the speedup of the pure Go implementation is even 20%:
```
CurlTransaction     3.87ms ± 4%  3.05ms ± 5%  -21.25%
CurlHash             140µs ± 8%   115µs ± 7%  -18.31%
```
## Type of change

The existing test suite has been extended to test the batched Curl also against the regular Curl golden test vectors as single input.

## Change checklist

### DIFF
--- a/curl/bct/curl.go
+++ b/curl/bct/curl.go
@@ -124,7 +124,7 @@ func (c *Curl) out(dst trinary.Trits, idx uint) {
 // transform transforms the sponge.
 func (c *Curl) transform() {
 	var ltmp, htmp [curl.StateSize]uint
-	transform(&ltmp, &htmp, &c.l, &c.h, curl.NumRounds)
+	transform(&ltmp, &htmp, &c.l, &c.h)
 	c.l, c.h = ltmp, htmp
 }
 
--- a/curl/bct/curl_test.go
+++ b/curl/bct/curl_test.go
@@ -1,6 +1,10 @@
 package bct_test
 
 import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
 	"strings"
 
 	"github.com/iotaledger/iota.go/consts"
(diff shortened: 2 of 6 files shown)


Wollac:
With this PR the performance improvements of the `amd64` Assembly version have decreased further. It has gone done to less than 8% when comparing Assembly to pure pure Go:
```
CurlTransaction     3.05ms ± 5%  2.86ms ± 5%  -6.28%
CurlHash             115µs ± 7%   106µs ± 6%  -7.65%
```
This introduces the question what speed-up still justifies having very hard to maintain and platform dependent assembly code for `amd64`. Or should it maybe removed, as this is also the only package in `iota.go` that includes Assembly code.

luca-moser:
If the new Go impl. is faster than the assembly code, then there is no need to longer have it in the lib.

Wollac:
> If the new Go impl. is faster than the assembly code, then there is no need to longer have it in the lib.

The assembly version is still 4% - 5% faster than the pure go version. The following compares `master` against the pure go verstion from `Wollac:curl-indices`:
```
CurlTransaction-48  3.18ms ±10%  3.05ms ± 5%  -4.28%  (p=0.007 n=16+16)
CurlHash-48          115µs ± 4%   115µs ± 7%    ~     (p=0.654 n=15+16)
```
However, this is really not that much and the pure go version from `Wollac:curl-indices` is still 6%-7% faster than the previous assembly version. That's exactly why I have been wondering whether keeping and maintaining an `amd64` assembly version still makes sense for a mere 4% speedup...

luca-moser:
Okay, now I got. Given the fact that the implementation shouldn't really change in the future, I don't think it is really necessary to remove it, even though you're probably the only person who understands the assembly version. I'd leave it in.

PR #278, opened 2021-04-08
Change "PoW" to "Pow" in the milestone json tags

muXxer:
To satisfy the bee team.


muXxer:
Okay we won't change it, lel

PR #288, opened 2021-05-19
Add ledgerIndex field to NodeOutputResponse

muXxer:


luca-moser:
I believe we wanted to add this also to `AddressBalanceResponse`

muXxer:
@luca-moser you already did, no? #283

PR #296, opened 2021-11-25
fix: nil pointer access in NewSemValiContextWorkingSet

dessaya:
# Description of change

fix: nil pointer access in `NewSemValiContextWorkingSet`

## Type of change

Just running `NewSemValiContextWorkingSet` would panic. This change fixes that.
(I'm using the `iota.go` code as a library from outside).


luca-moser:
Thanks for the PR. This is already fixed on branch `v3` at `eae1dcd8e4b08422a74a6a2cbba5c89e60ef6282`

Issue #298, opened 2021-12-11
Incorrect semantical validation of ChainConstrainedAddress 


lunfardo314:
In https://github.com/iotaledger/iota.go/blob/914ff5853384896ec13987aef511ab60c3623803/transaction.go#L470
the fragment `!referentialUnlockBlock.Chainable()` is not needed because it always `true`. 
This leads to impossibility to unlock extended address by alias output.
Example of failing test: https://github.com/iotaledger/wasp/blob/060ce0dc23997fb351c9f8e6d01c57221aa2763a/packages/transaction/tx_test.go#L164

luca-moser:
> the fragment !referentialUnlockBlock.Chainable() is not needed because it always true.

There are 3 implementations of `ReferentialUnlockBlock`, 2 of which return `true` for `Chainable`. This check is there to ensure that indeed for an input with a `ChainConstrainedAddress`, the corresponding unlock block is an `Alias`/`NFTUnlockBlock`.

The bug you are facing is because you're using the `TransactionBuilder` which has not been updated to correctly use the new `UnlockBlocks`, hence you are generating txs with `ReferenceUnlockBlocks` at positions where you must have a `Alias/NFTUnlockBlock`.

Looking at your test code, you are consuming the alias, unlock the state controller and the builder will place a `ReferenceUnlockBlock` instead of the required `AliasUnlockBlock`, hence your test doesn't pass.

I will therefore close your PR and keep this issue open until the builder has been updated.

luca-moser:
Will be tracked via https://github.com/iotaledger/iota.go/issues/379

PR #299, opened 2021-12-12
isc vm use case test failing

lunfardo314:
A test covering a bug in tx semantical validation on the main use case when alias output consumes. See issue https://github.com/iotaledger/iota.go/issues/298
The test fail because semantical validation code changes do not fixed it completely.


luca-moser:
See https://github.com/iotaledger/iota.go/issues/298#issuecomment-992228838

PR #320, opened 2022-03-01
Refactored number formatting in JSON api (hex and strings)

alexsporn:
- Centralized the hex string encoding/decoding into 2 public functions (`EncodeHex`, `DecodeHex`)
- Switched from using `encoding/hex` to use the `github.com/ethereum/go-ethereum/common/hexutil` pkg to enforce `0x` prefix usage
- Added helpers to encode/decode uint64 numbers (`EncodeUint64`, `DecodeUint64`)
- Encoding output amount fields as strings now

### DIFF
--- a/address_alias.go
+++ b/address_alias.go
@@ -1,7 +1,6 @@
 package iotago
 
 import (
-	"encoding/hex"
 	"encoding/json"
 	"fmt"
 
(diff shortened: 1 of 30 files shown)


luca-moser (approved):
Thanks for the PR, lgtm. I'll adjust the `uint256` with a subsequent PR to also use `0x` formatting.

Side note: the uint64s converted via `EncodeUint64` for API routes are kinda odd as the template strings could simply use `%d` instead of `%s`.

PR #322, opened 2022-03-01
Use newer native token validation rules with minted/melted fields 

luca-moser:


output_foundry.go
```diff
@@ -236,20 +239,25 @@ func (f *FoundryOutput) ValidateStateTransition(transType ChainTransitionType, n
...
 	default:
 		panic("unknown chain transition type in FoundryOutput")
 	}
 }
 
-func (f *FoundryOutput) checkStateGenesisTransition(semValCtx *SemanticValidationContext, thisFoundryID FoundryID, inSums NativeTokenSum, outSums NativeTokenSum) error {
+func (f *FoundryOutput) checkStateGenesisTransition(semValCtx *SemanticValidationContext, thisFoundryID FoundryID, outSums NativeTokenSum) error {
+	// minted and melted counter must be zero at genesis
+	switch {
+	case f.MintedTokens.Cmp(common.Big0) != 0:
```
lzpap, line 255: I'm not sure this is right. According to the specs, it should be possible to mint tokens in the tx which creates the foundry, therefore MintedTokens doesn't have to be 0 at genesis. But you can't melt tokens in the same tx, because there can never be already existing tokens on the input side, so TokenDiff will never be negative for the genesis transaction of a foundry.

output_foundry.go
```diff
@@ -333,11 +343,13 @@ func (f *FoundryOutput) checkStateChangeTransition(next ChainConstrainedOutput,
 		return fmt.Errorf("%w: maximum supply mismatch wanted %s but got %s", ErrInvalidChainStateTransition, f.MaximumSupply, nextState.MaximumSupply)
 	case f.TokenTag != nextState.TokenTag:
 		return fmt.Errorf("%w: token tag mismatch wanted %s but got %s", ErrInvalidChainStateTransition, f.TokenTag, nextState.TokenTag)
+	case f.MintedTokens.Cmp(nextState.MintedTokens) == 1:
+		return fmt.Errorf("%w: current minted supply (%s) bigger than next minted supply (%s)", ErrInvalidChainStateTransition, f.MintedTokens, nextState.MintedTokens)
+	case f.MeltedTokens.Cmp(nextState.MeltedTokens) == 1:
+		return fmt.Errorf("%w: current melted supply (%s) bigger than next melted supply (%s)", ErrInvalidChainStateTransition, f.MeltedTokens, nextState.MeltedTokens)
```
lzpap, line 349: Why isn't the foundry allowed to mint/melt in a state transition?

luca-moser: These lines mean that the current state counters can not be bigger than the next state counters.

output_foundry.go
luca-moser: Perhaps I am misreading the TIP, but given this rule https://github.com/iotaledger/tips/pull/38/files#diff-39a438c54a901ee9f1c75be884c6b90b8d2794dc799d4fd290eee3af4e28b8c6R2985 I read it as "when the foundry is new (`Current State` is empty), then `Minted` and `Melted` tokens must be 0. Did I misinterpret the TIP?3

lzpap: It might be hard to read, but it is defined above, that:
> - The **current state of the foundry** with `Foundry ID` `X` in a transaction is defined as the consumed foundry output
  where `Foundry ID` = `X`.
> - The **next state of the foundry** with `Foundry ID` `X` in a transaction is defined as the created foundry output
  where `Foundry ID` = `X`.

And that:
| A transaction that...   | Current State           | Next State               |
|-------------------------|-------------------------|--------------------------|
| Creates the foundry     | Empty                   | Output with `Foundry ID` |
| Transitions the foundry | Input with `Foundry ID` | Output with `Foundry ID` |
| Destroys the foundry    | Input with `Foundry ID` | Empty                    |


When the foundry is created, it does not appear as a consumed output in the trasanction, hence `Current State` is `empty`. `Next State` is the actual foundry output created in that transaction.

The line you referenced means that when there is no foundry output on the input side (current state empty), assume in the calculations that `Current State(Minted) = 0` and `Current State(Melted)=0`.

But it doesn't mean that `Next State(Minted)`, the actual field of the created foundry, must be `0`.

output_foundry.go
lzpap: Ok, that is reasonable.

output_foundry.go
luca-moser: Alright, fair enough. I was already wondering why this criteria suddenly came into play.

output.go
```diff
@@ -829,9 +829,10 @@ func OutputsSyntacticalFoundry() OutputsSyntacticalValidationFunc {
 			return fmt.Errorf("%w: output %d, less than equal zero", ErrFoundryOutputInvalidMaximumSupply, index)
 		}
 
-		if r := foundryOutput.CirculatingSupply.Cmp(foundryOutput.MaximumSupply); r == 1 {
-			return fmt.Errorf("%w: output %d, bigger than maximum supply", ErrFoundryOutputInvalidCirculatingSupply, index)
-		}
+		// TODO: more syntactic checks are possible but not defined in the TIP:
+		// - minted - melted > 0: foundry can never have melted more than minted
+		// - minted - melted <= max supply: can never have minted more than max supply
+		// - melted <= minted: can never have melted more than minted
```
lzpap, line 835: Where are the syntactic checks from the TIP?
 - Minted Tokens must not be greater than Maximum Supply.
 - Melted Tokens must not be greater than Minted Tokens.

output.go
```diff
@@ -829,9 +829,10 @@ func OutputsSyntacticalFoundry() OutputsSyntacticalValidationFunc {
 			return fmt.Errorf("%w: output %d, less than equal zero", ErrFoundryOutputInvalidMaximumSupply, index)
 		}
 
-		if r := foundryOutput.CirculatingSupply.Cmp(foundryOutput.MaximumSupply); r == 1 {
-			return fmt.Errorf("%w: output %d, bigger than maximum supply", ErrFoundryOutputInvalidCirculatingSupply, index)
-		}
+		// TODO: more syntactic checks are possible but not defined in the TIP:
+		// - minted - melted > 0: foundry can never have melted more than minted
+		// - minted - melted <= max supply: can never have minted more than max supply
```
lzpap, line 834: Do you think we should add these to the TIP?

PR #325, opened 2022-03-09
v3 NodeEventAPIClient improvements

alexsporn:
- Added new `NodeEventAPIClientSubscription` that gets returned together with the event channels.
- This object holds any errors that occurred while subscribing and allows to cleanly unsubscribe from events.

### DIFF
--- a/x/node_event_api_client.go
+++ b/x/node_event_api_client.go
@@ -49,6 +49,9 @@ const (
 var (
 	// ErrNodeEventAPIClientInactive gets returned when a NodeEventAPIClient is inactive.
 	ErrNodeEventAPIClientInactive = errors.New("node event api client is inactive")
+
+	// ErrNodeEventAPIClientSubscriptionAlreadyClosed gets returned when a NodeEventAPISubscription was already closed.
+	ErrNodeEventAPIClientSubscriptionAlreadyClosed = errors.New("node event api client subscription is already closed")
 )
 
 // NodeEventUnlockCondition denotes the different unlock conditions.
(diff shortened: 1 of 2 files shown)


x/node_event_api_client.go
```diff
@@ -97,6 +100,48 @@ type NodeEventAPIClient struct {
 	Errors chan error
 }
 
+// NodeEventAPIClientSubscription holds any error that happened when trying to subscribe to an events.
```
muXxer, line 103:
```suggestion
// NodeEventAPIClientSubscription holds any error that happened when trying to subscribe to an event.
```

x/node_event_api_client.go
```diff
@@ -97,6 +100,48 @@ type NodeEventAPIClient struct {
...
+}
+
+// Error holds any error that happened when trying to subscribe.
+func (s *NodeEventAPIClientSubscription) Error() error {
+	return s.error
+}
+
+// Close allows to close the subscription to cleanly unsubscribe from the node.
+// It returns ErrNodeEventAPIClientSubscriptionAlreadyClosed if called multiple times.
+func (s *NodeEventAPIClientSubscription) Close() error {
```
muXxer, line 131: Could it maybe be necessary to add some mutexes? In case clients use it in a multithreaded way?

alexsporn: I thought about it, but nothing in the lib is thread-safe per se, so I skipped it for this.
I checked and I could remove this error and just let the user call Close multiple times. The underlying lib won't complain if you unsubscribe more than once

muXxer: Yeah, sounds like a good alternative.

muXxer (approved):

luca-moser (approved):

PR #340, opened 2022-04-17
Add Accept Header MIME type

muXxer:


nodeclient/http.go
```diff
@@ -135,6 +146,10 @@ func do(httpClient *http.Client, baseURL string, ctx context.Context, userInfo *
 		req.URL.User = userInfo
 	}
 
+	if requestHeaderHook != nil {
+		requestHeaderHook(req.Header)
+	}
+
```
alexsporn, line 152: This should be moved to after the next block where the content-type was set, because you are modifying the headers after the hook again

nodeclient/http_api_client.go
```diff
@@ -324,28 +342,14 @@ func (client *Client) MessageMetadataByMessageID(ctx context.Context, msgID iota
...
-}
-
 // MessageByMessageID get a message by its message ID from the node.
 func (client *Client) MessageByMessageID(ctx context.Context, msgID iotago.MessageID, deSeriParas *iotago.DeSerializationParameters) (*iotago.Message, error) {
-	query := fmt.Sprintf(RouteMessageBytes, iotago.EncodeHex(msgID[:]))
+	query := fmt.Sprintf(RouteMessage, iotago.EncodeHex(msgID[:]))
 
 	res := &RawDataEnvelope{}
-	if _, err := client.Do(ctx, http.MethodGet, query, nil, res); err != nil {
+	if _, err := client.DoWithRequestHeaderHook(ctx, http.MethodGet, query, func(header http.Header) {
```
alexsporn, line 350: Maybe define a default hook that sets the accept header?

nodeclient/http_api_client.go
```diff
@@ -391,7 +397,9 @@ func (client *Client) OutputByID(ctx context.Context, outputID iotago.OutputID)
 	query := fmt.Sprintf(RouteOutput, outputID.ToHex())
 
 	res := &OutputResponse{}
-	if _, err := client.Do(ctx, http.MethodGet, query, nil, res); err != nil {
+	if _, err := client.DoWithRequestHeaderHook(ctx, http.MethodGet, query, func(header http.Header) {
+		header.Set("Accept", "application/json")
```
alexsporn, line 401: define and use `MIMEApplicationJSON = "application/json"` since you are also mentioning it in the comments with that name

alexsporn (approved):

PR #343, opened 2022-04-21
Refactors total supply logic

luca-moser:
The library used a `TotalSupply` constant reflecting the total supply of IOTA tokens in various validation logic. However, the total supply depends on the actual network in which the library is used. Therefore, the total supply is to be defined within the `ProtocolParameters` struct (renamed from `DeSeriParameters` to better reflect what it is).

The `TotalSupply` constant has been renamed to `TotalSupplyIOTA` as to make it more clear that it reflects the total supply on the IOTA mainnet. When there is a clear plan to make IOTA mainnet inflationary, this parameter is subject to be removed completely.


rent.go
```diff
@@ -22,7 +22,7 @@ var (
 
 	// ZeroRentParas are test parameters for de/serialization using zero vbyte rent cost.
 	// Only use this var in testing. Do not modify.
-	ZeroRentParas = &DeSerializationParameters{RentStructure: &RentStructure{
+	ZeroRentParas = &ProtocolParameters{RentStructure: &RentStructure{
```
muXxer, line 25: Isn't the total supply missing here?

muXxer: Maybe we need ZeroRentParasIOTA here?

muXxer (changes requested):

rent.go
alexsporn: I was thinking the same, but we don't run validation where we use ZeroRentParas, so it being 0 is ok

builder/message_builder.go
```diff
@@ -3,18 +3,17 @@ package builder
...
 	"github.com/iotaledger/iota.go/v3/pow"
 )
 
 // NewMessageBuilder creates a new MessageBuilder.
-func NewMessageBuilder() *MessageBuilder {
+func NewMessageBuilder(protoVersion byte) *MessageBuilder {
 	return &MessageBuilder{
 		msg: &iotago.Message{
-			ProtocolVersion: iotago.ProtocolVersion,
+			ProtocolVersion: protoVersion,
```
muXxer, line 16: Is this change necessary?
The message builder should always have the same protocol version like the created message.

builder/transaction_builder.go
```diff
@@ -89,11 +89,11 @@ func (b *TransactionBuilder) BuildAndSwapToMessageBuilder(deSeriParas *iotago.De
...
 
 // Build sings the inputs with the given signer and returns the built payload.
-func (b *TransactionBuilder) Build(deSeriParas *iotago.DeSerializationParameters, signer iotago.AddressSigner) (*iotago.Transaction, error) {
+func (b *TransactionBuilder) Build(protoParas *iotago.ProtocolParameters, signer iotago.AddressSigner) (*iotago.Transaction, error) {
 	switch {
 	case b.occurredBuildErr != nil:
 		return nil, b.occurredBuildErr
-	case deSeriParas == nil:
+	case protoParas == nil:
 		return nil, fmt.Errorf("%w: must supply de/serialization parameters", ErrTransactionBuilder)
```
muXxer, line 97:
```suggestion
		return nil, fmt.Errorf("%w: must supply protocol parameters", ErrTransactionBuilder)
```

doc.go
```diff
@@ -42,7 +42,7 @@
 //		}
 //
 //		// get the parameters for the storage deposit
-//		nodeRentParas := &iotago.DeSerializationParameters{RentStructure: &info.Protocol.RentStructure}
+//		nodeRentParas := &iotago.ProtocolParameters{RentStructure: &info.Protocol.RentStructure}
```
muXxer, line 45: Total supply missing here.

builder/message_builder.go
luca-moser: Yes, because the library might support multiple versions at the same time, there isn't a single lib version per protocol version.

PR #354, opened 2022-04-29
Check network id during tx essence syntactical checks

luca-moser:
Because now we can given `ProtocolParameters`.


tpkg/test_consts.go
```diff
@@ -5,14 +5,21 @@ import "github.com/iotaledger/iota.go/v3"
...
 	RentStructure: iotago.RentStructure{
 		VByteCost:    0,
 		VBFactorData: 0,
 		VBFactorKey:  0,
 	},
+	TokenSupply: TestTokenSupply,
 }
 
+// TestNetworkID is a test network ID.
+var TestNetworkID = iotago.NetworkIDFromString(TestProtoParas.NetworkName)
```
alexsporn, line 21: you can call TestProtoParas.Network()

PR #362, opened 2022-05-13
Renames Message to Block

luca-moser:
At the instances where we talk about `Message` as the core protocol model but not where it is used as a raw json message or when as an input to cryptographic functions.

We do this to better align to IOTA 2.0 naming


block.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	*m = *seri.(*Block)
+	return nil
+}
+
+// jsonBlock defines the JSON representation of a Block.
+type jsonBlock struct {
+	ProtocolVersion int `json:"protocolVersion"`
+	// The hex encoded IDs of the referenced parent blocks.
+	Parents []string `json:"parentBlockIds"`
```
alexsporn, line 231: do we maybe want to just call it parents in JSON too?

milestone.go
```diff
@@ -491,7 +491,7 @@ type jsonMilestone struct {
 	Timestamp           int                `json:"timestamp"`
 	ProtocolVersion     int                `json:"protocolVersion"`
 	PreviousMilestoneID string             `json:"previousMilestoneId"`
-	Parents             []string           `json:"parentMessageIds"`
+	Parents             []string           `json:"parentBlockIds"`
```
alexsporn, line 494: Here again, maybe just call it `parents`?

block.go
luca-moser: sounds good to me

thibault-martinez: Should we do the same for the API block metadata response ?

alexsporn: I renamed them

PR #364, opened 2022-05-13
Renames FeatureBlock to Feature

luca-moser:


output_basic.go
```diff
@@ -79,32 +79,32 @@ func BasicOutputUnlockConditionsArrayRules() serializer.ArrayRules {
...
 	// The amount of IOTA tokens held by the output.
 	Amount uint64
 	// The native tokens held by the output.
 	NativeTokens NativeTokens
 	// The unlock conditions on this output.
 	Conditions UnlockConditions
-	// The feature blocks which extending the output metadata.
-	Blocks FeatureBlocks
+	// The features on the output.
+	Feats Features
```
alexsporn, line 99: Call it `Features` because a feat is something else

luca-moser: The reason I've done this is because of the naming clash with the interface `Output`'s `Features()` function. However, I changed these interface functions to return sets (as one would want to get the set anyways), therefore adjusted the names and reverted back the fields within the output structs to `Features` etc.

doc.go
```diff
@@ -1,6 +1,6 @@
 // Package iotago provides IOTA data models, a node API client and builders to craft blocks and transactions.
 //
-// Creating Blocks:
+// Creating Features:
```
muXxer, line 3: Revert that change.

muXxer (approved):

muXxer (approved):

PR #365, opened 2022-05-17
Use distinct types for ID types

luca-moser:
Makes `BlockID` and `BlockIDs` actual types instead of aliases. This allows us to add methods to these types for hex conversion etc.

Due to the serializer's shortcoming, I've added some variable declarations as to make it work with the current serializer API. Given that we will be getting rid of it anyway, this seems like an ok solution for now.

Also added ToHex() to `ChainID` thus to `AliasID`/`FoundryID`/`NFTID`.

### DIFF
--- a/block.go
+++ b/block.go
@@ -1,9 +1,11 @@
 package iotago
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sort"
 
 	"golang.org/x/crypto/blake2b"
 
@@ -25,6 +27,9 @@ const (
 var (
 	// ErrBlockExceedsMaxSize gets returned when a serialized block exceeds BlockBinSerializedMaxSize.
 	ErrBlockExceedsMaxSize = errors.New("block exceeds max size")
+
+	// is an empty block ID
+	emptyBlockID = BlockID{}
 
 	blockPayloadGuard = serializer.SerializableGuard{
 		ReadGuard: func(ty uint32) (serializer.Serializable, error) {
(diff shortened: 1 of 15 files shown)


muXxer (approved):

milestone.go
```diff
@@ -108,16 +106,20 @@ type (
...
-	// MilestoneParentBlockID is a reference to a parent block.
-	MilestoneParentBlockID = BlockID
-	// MilestoneParentBlockIDs are references to parent blocks.
-	MilestoneParentBlockIDs = []MilestoneParentBlockID
 	// MilestoneMerkleProof is the merkle root within a milestone.
 	MilestoneMerkleProof = [MilestoneMerkleProofLength]byte
 )
 
+// MilestoneID is the ID of a Milestone.
+type MilestoneID [MilestoneIDLength]byte
```
alexsporn, line 114: Remove pointer from
`func (m *Milestone) ID() (*MilestoneID, error)`

block.go
```diff
@@ -65,32 +70,42 @@ func BlockParentArrayRules() serializer.ArrayRules {
...
-// BlockIDs are IDs of blocks.
-type BlockIDs = []BlockID
+func (b BlockID) MarshalBinary() (data []byte, err error) {
+	// copy
+	return b[:], nil
+}
 
-// BlockIDFromHexString converts the given block IDs from their hex to BlockID representation.
+// ToHex converts the given block ID to their hex representation.
+func (b *BlockID) ToHex() string {
```
alexsporn, line 87: why pointer method?

block.go
```diff
@@ -65,32 +70,42 @@ func BlockParentArrayRules() serializer.ArrayRules {
...
+	return EncodeHex(b[:])
+}
+
+// Empty tells whether the BlockID is empty.
+func (b *BlockID) Empty() bool {
```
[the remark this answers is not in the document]
luca-moser: Because it avoids a copy of the array.

block.go
luca-moser: See above, it avoids doing a copy of the array.

PR #386, opened 2022-06-21
Ported the merkle tree hasher from hornet to iota.go and added convenience functions to work with blockIDs

alexsporn:


merklehasher/inclusion_proof.go
```diff
@@ -0,0 +1,255 @@
...
+
+type leafValue struct {
+	Value []byte
+}
+
+type hashValue struct {
+	Value []byte
+}
+
+type InclusionProof struct {
```
Wollac, line 25: I would probable rename `InclusionProof` to just `Proof`. Usually it is called a "Merkle Proof" and that maybe makes it a bit more concise.
But "inclusion proof" is not wrong

Wollac (approved):

Wollac (approved):

PR #392, opened 2022-06-23
Move signingprovider from inx-coordinator to iota.go

muXxer:
We need the signing provider in several repositories (hornet, inx-coordinator, ...).

It uses a lot of types from iota.go, so it may make sense to directly move the code to iota.go.

What do you think? @luca-moser 

Needs to be merged after #391

### DIFF
--- a/signingprovider/signing_provider.go
+++ b/signingprovider/signing_provider.go
@@ -1,0 +1,144 @@
+package signingprovider
+
+import (
+	"crypto/ed25519"
+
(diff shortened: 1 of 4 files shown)


signingprovider/signing_provider.go
```diff
@@ -0,0 +1,144 @@
+package coordinator
```
alexsporn, line 1: Wrong pkg name

alexsporn (approved):

PR #406, opened 2022-08-22
Add query for native tokens to indexer client and fix alias query parameters

muXxer:
This PR adds the missing query parameters to ask the indexer for native token based filtering on all different output types.
It also adds the missing CreationParas based query to the alias outputs.

The Tag based query for aliases was removed, because there is no tag feature in alias outputs.


luca-moser (approved):
lgtm

PR #409, opened 2022-08-24
Check whether a referenced AliasOutput through an AliasUnlock is state transitioning

luca-moser:


lzpap (commented):
As far as I can tell, right now in the code:
 - If an alias is unlocked, its alias address is added to the unlocked idents.
 - If the alias address is referenced in any unlock condition in other inputs and the alias did not perform a state transition, validation fails. This is correct behavior.
 - However, if it is not referenced but rather the alias address is present in a sender or issuer feature on the output side, validation still passes. This is not correct behavior.

Recommendation: remove the alias address from unlocked idents if it wasn't state transitioned, even if it was not referenced in unlocks.

lzpap (approved):

transaction.go
```diff
@@ -480,31 +480,16 @@ func TxSemanticInputUnlocks() TxSemanticValidationFunc {
...
-
-			// there can be multiple entries given a chain output
-			aliasAddr, ok := v.Ident.(*AliasAddress)
-			if !ok {
-				continue
-			}
+				// for alias outputs which are not state transitioning, we do not add it to the set of unlocked chains
+				if current, ok := chainConstrOutput.(*AliasOutput); ok {
+					next, hasNextState := svCtx.WorkingSet.OutChains[chainID]
+					if !hasNextState || (current.StateIndex+1 != next.(*AliasOutput).StateIndex) {
```
muXxer, line 487: I was not able to do it, but do you think someone could create a malicious transaction somehow, where this chainID points to an NFTOutput instead, and cause a panic in the lib?

muXxer (approved):

transaction.go
muXxer: Funny, I think I created a test case where I have the exact same key twice in the OutChains map, but go still picks the right result from the map because one key seems to be an `aliasID` and the other one an `nftID`.

muXxer (approved):

PR #413, opened 2022-09-06
Relaxes max native token count to account only for distinct tokens

luca-moser:
According to the TIP, the max native token count should only count distinct native tokens instead of multiple occurrences of the same token.

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -736,10 +736,17 @@ func TxSemanticNativeTokens() TxSemanticValidationFunc {
 		if err != nil {
 			return fmt.Errorf("invalid output native token set: %w", err)
 		}
-		outNTCount := len(svCtx.WorkingSet.OutNativeTokens)
 
-		if inNTCount+outNTCount > MaxNativeTokensCount {
-			return fmt.Errorf("%w: native token count (in %d + out %d) exceeds max of %d", ErrMaxNativeTokensCountExceeded, inNTCount, outNTCount, MaxNativeTokensCount)
+		distinctNTCount := make(map[NativeTokenID]struct{})
+		for nt := range svCtx.WorkingSet.InNativeTokens {
+			distinctNTCount[nt] = struct{}{}
+		}
+		for nt := range svCtx.WorkingSet.OutNativeTokens {
+			distinctNTCount[nt] = struct{}{}
+		}
+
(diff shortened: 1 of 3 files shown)


muXxer (approved):

transaction.go
```diff
@@ -722,23 +722,31 @@ func TxSemanticNativeTokens() TxSemanticValidationFunc {
 	return func(svCtx *SemanticValidationContext) error {
 		// native token set creates handle overflows
 		var err error
-		var inNTCount, outNTCount int
-		svCtx.WorkingSet.InNativeTokens, inNTCount, err = svCtx.WorkingSet.Inputs.NativeTokenSum()
+		svCtx.WorkingSet.InNativeTokens, err = svCtx.WorkingSet.Inputs.NativeTokenSum()
 		if err != nil {
 			return fmt.Errorf("invalid input native token set: %w", err)
 		}
+		inNTCount := len(svCtx.WorkingSet.InNativeTokens)
```
lzpap, line 729: Nature loves symmetry, so let's have the same check for `outNTCount` before calculating `distinctNativeCount`.

luca-moser: The transaction would not pass syntactical validation if the count would exceed the distinct native count beforehand.

lzpap: Yes, that's clear. I wast just suggesting this optimization in order not to compute `distinctNTCount` when `len(svCtx.WorkingSet.OutNativeTokens)` would already exceed the limit.

luca-moser: I do not understand the optimisation: `len(svCtx.WorkingSet.OutNativeTokens)` can not exceed `Max Native Tokens Count` here as the transaction would already fail at syntactical validation otherwise, therefore the check will never be true.

On the other hand for inputs this check makes sense as the inputs are only available at semantical validation.

lzpap: Good point. I just realized it is indeed checked in https://github.com/iotaledger/iota.go/blob/2f37dd013dee53077201133ae74aa67559cdfdc0/output.go#L748 Let me add this check to TIP-20 because it is not present in `Transaction Syntactical Validation Rules`. According to TIP-20, native token counts are checked per output (syntactic) and then in the semantic tx validation accross the all transaction, but not for only outputs.

PR #419, opened 2023-01-19
Add ByteSizeKey and ByteSizeData to NonEphemeralObject

muXxer:
This PR adds some ByteSize functions to the `NonEphemeralObject` interface to get the actual used amount of bytes for an object.


muXxer:
Never gonna give you up, never gonna let you down..... but maybe in this case, lets close it.

PR #430, opened 2023-05-24
Update attestation comparison

jonastheis:
# Description of change

This PR updates `Compare` of `Attestation` to take the slot index of it as the main factor for comparison. 
If equal, the lower issuing time of the carrying block is used. This is to prevent frequent switching of attestations in more recent blocks.

### DIFF
--- a/attestation.go
+++ b/attestation.go
@@ -42,9 +42,9 @@ func (a *Attestation) Compare(other *Attestation) int {
 		return 1
 	case a.SlotCommitmentID.Index() < other.SlotCommitmentID.Index():
 		return -1
-	case a.IssuingTime.Before(other.IssuingTime):
-		return 1
 	case a.IssuingTime.After(other.IssuingTime):
+		return 1
+	case a.IssuingTime.Before(other.IssuingTime):
 		return -1
 	default:
 		return bytes.Compare(a.BlockContentHash[:], other.BlockContentHash[:])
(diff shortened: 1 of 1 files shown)


attestation.go
```diff
@@ -38,9 +38,13 @@ func (a *Attestation) Compare(other *Attestation) int {
 		return -1
 	case other == nil:
 		return 1
-	case a.IssuingTime.After(other.IssuingTime):
+	case a.SlotCommitmentID.Index() > other.SlotCommitmentID.Index():
+		return 1
+	case a.SlotCommitmentID.Index() < other.SlotCommitmentID.Index():
+		return -1
+	case a.IssuingTime.Before(other.IssuingTime):
```
karimodm: In this condition you return 1 if `a` is before than `other` on issuingTime, but above you return -1 if `a`'s `SlotCommitmentID.Index()` is before `other`'s commitment index.
The role of this method is for you to evaluate if you should refresh an attestation or not.
As far as I see these checks should be exactly reversed: if `other` commits to an higher index than `a` you return 1, but to avoid refreshing too often I would add additional checks:
1. If the `other`'s block Index is greater than you return 1: you refresh the attestation because you have a block inside a newer slot than the one you previously had.
2. If the IssuingTime of other is greater you return -1. This is exactly to avoid to refresh attestations too often, and always take the earliest block contained in each epoch that also commit to the same slot.

karimodm (approved):

PR #431, opened 2023-06-02
Align client lib to API spec

jkrvivian:
# Description of change

As title describe.

## Type of change

- Breaking change

## Change checklist

### DIFF
--- a/nodeclient/core_models.go
+++ b/nodeclient/core_models.go
@@ -11,10 +11,7 @@ import (
 var _internalAPI = iotago.V3API(&iotago.ProtocolParameters{})
 
 type (
-	httpOutput   interface{ iotago.Output }
-	outputResAux struct {
-		Output httpOutput `serix:"0,mapKey=output"`
-	}
+	httpOutput interface{ iotago.Output }
 )
 
 func must(err error) {
@@ -32,6 +29,8 @@ func init() {
 }
 
 type (
+	Versions []uint32
+
 	// RoutesResponse defines the response of a GET routes REST API call.
 	RoutesResponse struct {
 		Routes []string `json:"routes"`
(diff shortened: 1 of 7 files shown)


nodeclient/core_models.go
```diff
@@ -41,40 +40,56 @@ type (
 	InfoResponse struct {
 		// The name of the node software.
 		Name string `json:"name"`
-		// The semver version of the node software.
+		// The server version of the node software.
```
piotrm50: This should stay `semver`, for `semantic versioning`. 

```suggestion
		// The semver version of the node software.
```

nodeclient/core_models.go
```diff
@@ -93,113 +108,82 @@ type (
...
+		// TxStateReason if applicable indicates the error that occurred during the transaction processing.
+		TxStateReason string `json:"txStateReason,omitempty"`
+		// ReissuePayload whether the block should be reissued.
+		ReissuePayload *bool `json:"reissuePayload,omitempty"`
 	}
 
 	// OutputMetadataResponse defines the response of a GET outputs metadata REST API call.
 	OutputMetadataResponse struct {
-		// The hex encoded ID of the block.
+		// BlockID is the block ID that contains the output.
```
piotrm50, line 149: What happens if there are multiple included attachments of the same transactions?

nodeclient/core_models.go
```diff
@@ -93,113 +108,82 @@ type (
...
-	TreasuryResponse struct {
-		MilestoneID string `json:"milestoneId"`
-		Amount      string `json:"amount"`
-	}
-
-	// ReceiptsResponse defines the response for receipts GET related REST API calls.
-	ReceiptsResponse struct {
-		Receipts []*ReceiptTuple `json:"receipts"`
+		// IncludedCommitmentID is the commitment ID that includes the output.
+		IncludedCommitmentID string `json:"includedCommitmentId,omitempty"`
```
piotrm50, line 162: What happens if an output is not committed yet and a client requests it? Is EmptyCommittment returned?

nodeclient/core_models.go
```diff
@@ -93,113 +108,82 @@ type (
...
+		// The cumulative weight of the requested slot.
+		CumulativeWeight uint64 `json:"cumulativeWeight"`
+		// TODO: decide what else to add here.
 	}
 
-	// CommitmentUTXOChangesResponse defines the response of a GET milestone UTXO changes REST API call.
-	CommitmentUTXOChangesResponse struct {
-		// The index of the milestone.
+	// SlotUTXOResponse defines the response for UTXO slot REST API call.
+	SlotUTXOResponse struct {
```
piotrm50: In Swagger this is called `UTXOChangesResponse`. I think we should keep the naming consistent between code and specs.

nodeclient/core_models.go
```diff
@@ -93,113 +108,82 @@ type (
...
+		// LatestCommitmentID is the latest commitment ID of a node.
+		LatestCommitmentID string `json:"latestCommitmentId"`
 	}
 
-	// ReceiptTuple represents a receipt and the milestone index in which it was contained.
-	ReceiptTuple struct {
-		Receipt        *json.RawMessage `json:"receipt"`
-		MilestoneIndex iotago.SlotIndex `json:"milestoneIndex"`
+	// CommitmentUTXOChangesResponse defines the response of a GET milestone UTXO changes REST API call.
+	CommitmentInfoResponse struct {
```
piotrm50: This one is called `Commitment` in specs.

nodeclient/core_models.go
```diff
@@ -93,113 +108,82 @@ type (
...
-		// The ratio of referenced blocks in relation to new blocks of the last confirmed milestone.
-		ReferencedRate float64 `json:"referencedRate"`
-	}
-
-	// TipsResponse defines the response of a GET tips REST API call.
-	TipsResponse struct {
-		// The hex encoded block IDs of the tips.
-		TipsHex []string `json:"tips"`
+	// BlockIssuanceResponse defines the response of a GET block issuance REST API call.
+	BlockIssuanceResponse struct {
```
piotrm50, line 112: This one is called `IssuanceBlockHeaderResponse` in specs.

piotrm50 (changes requested):
I think we should keep the naming between API spec and code consistent. I commented the ones that I found were not aligned.

nodeclient/core_models.go
jkrvivian: We returned the earliest block that got confirmed.

nodeclient/core_models.go
jkrvivian: Then it would not be in the response, that's why it's omitempty. Same as CommitmentIDSpent and TransactionIDSpent.

nodeclient/core_models.go
jkrvivian: Actually, we should have a CommitmentInfoResponse in spec bc we might want to add more info in this response, e.g., all roots of trees for proof of inclusion.

nodeclient/core_models.go
jkrvivian: I like `BlockIssuanceResponse` better, also fits well with the API route. :grin: 
So changes will be made in spec.

piotrm50 (approved):

PR #432, opened 2023-06-05
Replace `time.Time` with `uint64` as block timestamp

PhilippGackstatter:
# Description of change

Replace `time.Time` with `uint64` on the `Block` type. The timestamp is supposed to be encoded as `uint64` but `time.Time` is serialized as an `int64`. Out of the two options of 1) changing how `Time` is serialized and 2) replacing the type of the timestamp on the block, the latter was chosen. This option is both easier to implement and the convenience of working with `Time` can still be added via getters and setters.
The approach for converting between a Unix timestamp in `int64` and `uint64` and vice versa, is to truncate to the common value range both types can represent, which still allows for any dates to be represented between `1970` and `2262`.

## Type of change

Tests were added to ensure the correct conversion between the `uint64` we need and the expected `int64` nanosecond timestamp used by `time.Time`.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

## Change checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/attestation.go
+++ b/attestation.go
@@ -25,7 +25,7 @@ type Attestation struct {
 func NewAttestation(block *Block) *Attestation {
 	return &Attestation{
 		IssuerID:         block.IssuerID,
-		IssuingTime:      block.IssuingTime,
+		IssuingTime:      block.IssuingTime(),
 		SlotCommitmentID: block.SlotCommitment.MustID(),
 		BlockContentHash: lo.PanicOnErr(block.ContentHash()),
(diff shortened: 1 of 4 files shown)


PhilippGackstatter:
Ported to the serializer in hive.go. See iotaledger/hive.go#533 for details.

PR #444, opened 2023-06-25
Use Commitment input for time-related functionality

PhilippGackstatter:
# Description of change

1. Use the Commitment input for anything that requires a slot index during validation, i.e.
    - Timelock Unlock Condition
    - Expiration Unlock Condition
    - Staking Feature
    - Block Issuer Feature
    - Delegation Output
2. Implement voting power slot calculation according to the committee selection procedure. It is incomplete and is expected to be merged in that state for now.
3. Improve variable names to be more consistent with the rest of the code, e.g. `BIFeat` -> `BlockIssuerFeat`.
4. Change the staking feature validation to allow removing a staking feature only when the `End Epoch` has passed, instead of when the current epoch is equal `End Epoch`. The TIP was updated accordingly.
5. Change the default value of `epochNearingThreshold` in the protocol params to include MaxCA, as it is intended. This will probably be changed in the future again, but it was necessary to properly test the delegation changes.

## Type of change

Updated tests.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

## Change checklist


output_test.go
```diff
@@ -1110,7 +1164,7 @@ func TestAccountOutput_UnlockableBy(t *testing.T) {
 				if tt.identCanUnlockInstead == nil {
 					return
 				}
-				canUnlockInstead, err := tt.current.UnlockableBy(tt.identCanUnlockInstead, tt.next, tt.txCreationTime)
+				canUnlockInstead, err := tt.current.UnlockableBy(tt.identCanUnlockInstead, tt.next, tt.commitmentInputTime+tt.maxCommittableAge, tt.commitmentInputTime+tt.minCommittableAge)
```
muXxer, line 1167: Is there anywhere an explanation why it is like that?

I find it hard to understand why the `pastBoundedSlotIndex` is `commitmentInputTime+maxCommittableAge` and why the `futureBoundedSlotIndex` is `commitmentInputTime+minCommittableAge` :sweat_smile:

vm/vm.go
```diff
@@ -172,6 +172,20 @@ func TotalManaOut(outputs iotago.Outputs[iotago.TxEssenceOutput], allotments iot
...
 }
 
+// PastBoundedSlotIndex calculates the past bounded slot for the given slot.
+// Given any slot index of a commitment input, the result of this function is a slot index
+// that is at least equal to the slot of the block in which it was issued, or higher.
+func (params *Params) PastBoundedSlotIndex(slot iotago.SlotIndex) iotago.SlotIndex {
+	return slot + params.API.ProtocolParameters().MaxCommittableAge()
+}
+
+// FutureBoundedSlotIndex calculates the past bounded slot for the given slot.
```
muXxer: The first line of the comment seems wrong. Shouldn't it be future bounded slot?

vm/vm.go
```diff
@@ -172,6 +172,20 @@ func TotalManaOut(outputs iotago.Outputs[iotago.TxEssenceOutput], allotments iot
 	return totalOut, nil
 }
 
+// PastBoundedSlotIndex calculates the past bounded slot for the given slot.
+// Given any slot index of a commitment input, the result of this function is a slot index
+// that is at least equal to the slot of the block in which it was issued, or higher.
+func (params *Params) PastBoundedSlotIndex(slot iotago.SlotIndex) iotago.SlotIndex {
```
muXxer: Mhh it was hard for me to understand, until I saw that the commitment input slot is passed here.
Would it maybe make sense to rename `slot` to commitmentSlot` or `commitmentInputSlot` ?

vm/vm.go
```diff
@@ -397,12 +411,28 @@ func identToUnlock(vmParams *Params, input iotago.Output, inputIndex uint16) (io
...
-		return returnIdent
+func checkExpiration(vmParams *Params, output iotago.Output) (iotago.Address, error) {
+	if output.UnlockConditionSet().HasExpirationCondition() {
+		commitment := vmParams.WorkingSet.Commitment
+
+		if commitment == nil {
+			return nil, iotago.ErrExpirationConditionCommitmentInputRequired
+		}
+
+		futureBoundedSlotIndex := vmParams.FutureBoundedSlotIndex(commitment.Index)
```
muXxer, line 426: Random side question, should we rename `Index` to `Slot` maybe in the commitments?

vm/stardust/vm_stardust.go
```diff
@@ -133,20 +133,19 @@ func accountGenesisValid(current *iotago.AccountOutput, vmParams *vm.Params) err
...
+		if nextBlockIssuerFeat.ExpirySlot < vmParams.PastBoundedSlotIndex(vmParams.WorkingSet.Commitment.Index) {
 			return ierrors.Wrap(iotago.ErrInvalidBlockIssuerTransition, "block issuer feature expiry set too soon")
 		}
 	}
 
 	if stakingFeat := current.FeatureSet().Staking(); stakingFeat != nil {
 		if err := accountStakingGenesisValidation(current, stakingFeat, vmParams); err != nil {
-			// TODO: Change back.
-			return ierrors.Wrapf(iotago.ErrAccountMissing, "%w", err)
+			return ierrors.Wrapf(iotago.ErrInvalidStakingTransition, "%w", err)
```
muXxer: You can use Join here instead.

vm/stardust/vm_stardust.go
```diff
@@ -266,22 +265,21 @@ func accountBlockIssuerSTVF(input *vm.ChainOutputWithCreationTime, next *iotago.
...
-		if nextBIFeat.ExpirySlot != 0 && nextBIFeat.ExpirySlot != currentBIFeat.ExpirySlot && nextBIFeat.ExpirySlot < txSlotIndex+vmParams.API.ProtocolParameters().EvictionAge() {
+		if nextBlockIssuerFeat.ExpirySlot != currentBlockIssuerFeat.ExpirySlot && nextBlockIssuerFeat.ExpirySlot < vmParams.PastBoundedSlotIndex(commitmentInputIndex) {
 			return ierrors.Wrap(iotago.ErrInvalidBlockIssuerTransition, "block issuer feature expiry set too soon")
 		}
-
-	} else if nextBIFeat != nil {
+	} else if nextBlockIssuerFeat != nil {
 		// if the block issuer feature has expired, it must either be removed or expiry extended.
-		if nextBIFeat.ExpirySlot != 0 && nextBIFeat.ExpirySlot < txSlotIndex+vmParams.API.ProtocolParameters().EvictionAge() {
+		if nextBlockIssuerFeat.ExpirySlot < vmParams.PastBoundedSlotIndex(commitmentInputIndex) {
```
muXxer: Would it maybe make sense to calculate the `vmParams.PastBoundedSlotIndex(commitmentInputIndex` only once at the beginning of this function? It's used several times.

Maybe also the `vmParams.PastBoundedSlotIndex(commitmentInputIndex)+vmParams.API.ProtocolParameters().MaxCommittableAge())` ?

vm/stardust/vm_stardust.go
```diff
@@ -358,14 +356,22 @@ func accountStakingSTVF(chainID iotago.ChainID, current *iotago.AccountOutput, n
 	_, isClaiming := vmParams.WorkingSet.Rewards[chainID]
 
 	if currentStakingFeat != nil {
+		commitment := vmParams.WorkingSet.Commitment
+		if commitment == nil {
+			return ierrors.Wrapf(iotago.ErrInvalidStakingTransition, "%w", iotago.ErrInvalidStakingCommitmentInput)
```
muXxer: Use Join

vm/stardust/vm_stardust.go
```diff
@@ -472,32 +483,42 @@ func accountDestructionValid(input *vm.ChainOutputWithCreationTime, vmParams *vm
...
 	}
 
 	stakingFeat := outputToDestroy.FeatureSet().Staking()
 	if stakingFeat != nil {
-		_, isClaiming := vmParams.WorkingSet.Rewards[input.ChainID]
+		// This case should never occur as the staking feature requires the presence of a block issuer feature,
+		// which also requires a commitment input.
+		commitment := vmParams.WorkingSet.Commitment
+		if commitment == nil {
+			return fmt.Errorf("%w: %w", iotago.ErrInvalidStakingTransition, iotago.ErrInvalidStakingCommitmentInput)
```
muXxer: Please don't use fmt.Errorf at all, and here you can use `ierrors.Join`

vm/stardust/vm_stardust.go
```diff
@@ -472,32 +483,42 @@ func accountDestructionValid(input *vm.ChainOutputWithCreationTime, vmParams *vm
...
 	outputToDestroy := input.Output.(*iotago.AccountOutput)
 
-	BIFeat := outputToDestroy.FeatureSet().BlockIssuer()
-	if BIFeat != nil {
-		if BIFeat.ExpirySlot == 0 || BIFeat.ExpirySlot >= vmParams.WorkingSet.Tx.Essence.CreationTime {
-			// TODO: better error
+	blockIssuerFeat := outputToDestroy.FeatureSet().BlockIssuer()
+	if blockIssuerFeat != nil {
+		if vmParams.WorkingSet.Commitment == nil {
+			return fmt.Errorf("%w: block issuer feature validation requires a commitment input", iotago.ErrInvalidBlockIssuerTransition)
```
muXxer: Please don't use fmt.Errorf at all, and here you can use ierrors.Join

muXxer (changes requested):

vm/vm.go
PhilippGackstatter: Yeah, I'm also not super happy with `Index`. Maybe `SlotIndex` is most precise.

vm/stardust/vm_stardust.go
PhilippGackstatter: Agreed with the first point.

Yes, to the second point, but I don't know how to name that variable. Maybe leaving it anonymous by just calculating it as `pastBoundedSlotIndex+MCA` is better than a name that doesn't help? Not sure.

vm/vm.go
muXxer: Lets change it in another PR! :+1:

vm/stardust/vm_stardust.go
muXxer: Isn't it something like `maxCommitablePastBoundedSlot` :sweat_smile: ?

api_protocol_parameters.go
```diff
@@ -44,17 +44,18 @@ type basicProtocolParameters struct {
...
-	// when slots stop being a consumable accounts' state relative to the latest committed slot.
-	EvictionAge SlotIndex `serix:"16,mapKey=evictionAge"`
+	// MinCommittableAge is the minimum age relative to the accepted tangle time slot index that a slot can be committed.
+	MinCommittableAge SlotIndex `serix:"16,mapKey=minCommittableAge"`
+	// MaxCommittableAge is the maximum age for a slot commitment to be included in a block relative to the slot index of the block issuing time.
+	MaxCommittableAge SlotIndex `serix:"17,mapKey=maxCommittableAge"`
 	// LivenessThreshold is used by tip-selection to determine the if a block is eligible by evaluating issuingTimes
 	// and commitments in its past-cone to ATT and lastCommittedSlot respectively.
-	LivenessThreshold SlotIndex `serix:"17,mapKey=livenessThreshold"`
+	LivenessThreshold SlotIndex `serix:"18,mapKey=livenessThreshold"`
```
cyberphysic4l: I would like to change the order to reflect the constraints on the values here, livenessThreshold < MinCA < MaxCA < EpochNearing 
```suggestion
	// LivenessThreshold is used by tip-selection to determine the if a block is eligible by evaluating issuingTimes
	// and commitments in its past-cone to ATT and lastCommittedSlot respectively.
	LivenessThreshold SlotIndex `serix:"16,mapKey=livenessThreshold"`
    // MinCommittableAge is the minimum age relative to the accepted tangle time slot index that a slot can be committed.
	MinCommittableAge SlotIndex `serix:"17,mapKey=minCommittableAge"`
	// MaxCommittableAge is the maximum age for a slot commitment to be included in a block relative to the slot index of the block issuing time.
	MaxCommittableAge SlotIndex `serix:"18,mapKey=maxCommittableAge"`
```

cyberphysic4l (approved):

vm/stardust/vm_stardust.go
cyberphysic4l: How about minLockedSlotIndex as this value is the minimum future slot that we require an output to be locked until. So we will have <minLockedSlotIndex := pastBoundedSlotIndex + vmParams.API.ProtocolParameters().MaxCommittableAge()>

cyberphysic4l: I defined minManalockedSlotIndex for this.
74957cb14e0012ac23c7036e8f85418542d0a332

vm/vm.go
cyberphysic4l: I think we actually discussed removing the word index from everything. I think we can just call these Slot and Epoch and drop the word index everywhere.

jonastheis: Yep, let's do what @cyberphysic4l said in a separate PR in iota.go, iota-core and if necessary hive.go as well

muXxer (approved):

PR #445, opened 2023-06-27
Implement WorkScore

cyberphysic4l:
# Description of change

Implement the work score for blocks and their contents, including the transaction and all of its contents. 

The work score aims to approximate the computational requirements from nodes to process an object from end-to-end of the data flow.

For example, signature checks are compuationally expensive. Any object requiring a signature check to be carried out during processing will have a signature check component included in its work score and this will increase the overall work score of the object.


api.go
```diff
@@ -149,8 +151,8 @@ func (p ProtocolParameters) EpochDurationInSlots() SlotIndex {
...
-		p.Version, p.NetworkName, p.Bech32HRP, p.MinPoWScore, p.RentStructure, p.TokenSupply, p.GenesisUnixTimestamp, p.SlotDurationInSeconds, p.SlotsPerEpochExponent, p.ManaGenerationRate, p.ManaGenerationRateExponent, p.ManaDecayFactors, p.ManaDecayFactorsExponent, p.ManaDecayFactorEpochsSum, p.ManaDecayFactorEpochsSumExponent, p.StakingUnbondingPeriod, p.EvictionAge, p.LivenessThreshold, p.EpochNearingThreshold)
+	return fmt.Sprintf("ProtocolParameters: {\n\tVersion: %d\n\tNetwork Name: %s\n\tBech32 HRP Prefix: %s\n\tMinimum PoW Score: %d\n\tRent Structure: %v\n\tToken Supply: %d\n\tGenesis Unix Timestamp: %d\n\tSlot Duration in Seconds: %d\n\tSlots per Epoch Exponent: %d\n\tMana Generation Rate: %d\n\tMana Generation Rate Exponent: %d\t\nMana Decay Factors: %v\n\tMana Decay Factors Exponent: %d\n\tMana Decay Factor Epochs Sum: %d\n\tMana Decay Factor Epochs Sum Exponent: %d\n\tStaking Unbonding Period: %d\n\tEviction Age: %d\n\tLiveness Threshold: %d\n\tWorkScore Structure: %d\n\tEpoch Nearing Threshold: %d\n}",
```
muXxer: I guess the `%d` formatting is wrong in that context. Shouldn't it be `%v` or `%s` ?

workscore.go
```diff
@@ -0,0 +1,32 @@
+package iotago
+
+// WorkScore defines the type of work score used to denote the computation costs of processing an object.
+type WorkScore uint64
+
+type WorkScoreFactor byte
```
muXxer: Is 255 enough?
Maybe we should go for uint16 to be future proof?

workscore.go
```diff
@@ -0,0 +1,32 @@
...
+
+type WorkScoreStructure struct {
+	FactorData          WorkScoreFactor `serix:"0,mapKey=factorData"`
```
muXxer: Maybe add a section for factors and one for scores to the struct?
Then you don't need to repeat the "factor" or "workscore" in all the names.

allotment.go
```diff
@@ -21,6 +21,10 @@ func (a Allotments) Size() int {
 	return len(a) * (AccountIDLength + serializer.UInt64ByteSize)
 }
 
+func (a Allotments) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	return workScoreStructure.FactorAllotment.Multiply(a.Size())
```
muXxer: IMO we should take network traffic into account, so we should add a `workScoreStructure.FactorData.Multiply(a.Size()) + ` everywhere.

```suggestion
	return workScoreStructure.FactorData.Multiply(a.Size()) + workScoreStructure.FactorAllotment.Multiply(a.Size())
```

api.go
```diff
@@ -126,9 +126,11 @@ type ProtocolParameters struct {
 	// LivenessThreshold is used by tipselection to determine the if a block is eligibile by evaluating issuingTimes
 	// and commitments in its pastcone to ATT and lastCommittedSlot respectively.
 	LivenessThreshold SlotIndex `serix:"17,mapKey=liveNessThreshold"`
+	// WorkScoreStructure is a collection of work scores for different object types.
+	WorkScoreStructure WorkScoreStructure `serix:"18,mapKey=workScoreStructure"`
```
muXxer: Maybe we can move this structure below the rentstructure parameter (but maybe after the new protocol parameters PR got merged).

block.go
```diff
@@ -269,3 +270,19 @@ func (b *Block) DoPOW(ctx context.Context, targetScore float64) error {
 	b.Nonce = nonce
 	return nil
 }
+
+func (b *Block) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// Work Score for parents is the score for max parents plus a penalty for any less than max parents
+	numMissingParents := BlockMaxParents - (len(b.StrongParents) + len(b.WeakParents) + len(b.ShallowLikeParents))
```
muXxer: This is not correct.
`BlockMaxParents` is the maximum amount of parents per "parents field". So you could have a maximum of 24 parents here, which would produce a negative result here?

IMO we should do that a bit differently.
`WeakParents` and `ShallowLikeParents` should not be taken into account at all IMO.
For `StrongParents` we should punish for everything below 4 parents, but we shouldn't charge anything if the `StrongParents` are >= 4. What do you think?

```golang

workscore := 0
if len(b.StrongParents) < workScoreStructure.MinStrongParentsThreshold {
	workscore += workScoreStructure.FactorMissingParent.Multiply(workScoreStructure.MinStrongParentsThreshold - len(b.StrongParents))
}
```

workscore.go
```diff
@@ -0,0 +1,32 @@
...
+	FactorInput         WorkScoreFactor `serix:"1,mapKey=factorInput"`
+	FactorAllotment     WorkScoreFactor `serix:"2,mapKey=factorAllotment"`
+	FactorMissingParent WorkScoreFactor `serix:"3,mapKey=factorMissingParent"`
+
+	WorkScoreOutput           WorkScore `serix:"4,mapKey=workScoreOutput"`
+	WorkScoreStaking          WorkScore `serix:"5,mapKey=workScoreStaking"`
+	WorkScoreBlockIssuer      WorkScore `serix:"6,mapKey=workScoreBlockIssuer"`
+	WorkScoreEd25519Signature WorkScore `serix:"7,mapKey=workScoreEd25519Signature"`
+	WorkScoreNativeToken      WorkScore `serix:"8,mapKey=workScoreNativeToken"`
+	WorkScoreMaxParents       WorkScore `serix:"9,mapKey=workScoreMaxParents"`
```
muXxer: This can be removed.

unlock_nft.go
```diff
@@ -35,3 +35,7 @@ func (r *NFTUnlock) Type() UnlockType {
 func (r *NFTUnlock) Size() int {
 	return NFTUnlockSize
 }
+
+func (r *NFTUnlock) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	return workScoreStructure.WorkScoreEd25519Signature
```
muXxer: Do we really need to charge the workscore for the signature here?
Isn't that just a simple check if the signature unlock is there (which is already taken into account in the workscore?)

unlock_reference.go
```diff
@@ -35,3 +35,7 @@ func (r *ReferenceUnlock) Type() UnlockType {
 func (r *ReferenceUnlock) Size() int {
 	return ReferenceUnlockSize
 }
+
+func (r *ReferenceUnlock) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	return workScoreStructure.WorkScoreEd25519Signature
```
muXxer: Do we really need to charge the workscore for the signature here?
Isn't that just a simple check if the signature unlock is there (which is already taken into account in the workscore?)

unlock_cond_timelock.go
```diff
@@ -23,6 +23,11 @@ func (s *TimelockUnlockCondition) VBytes(rentStruct *RentStructure, _ VBytesFunc
 	return rentStruct.VBFactorData.Multiply(serializer.SmallTypeDenotationByteSize + serializer.UInt64ByteSize)
 }
 
+func (s *TimelockUnlockCondition) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// TimelockUnlockCondition requires a signature check, but on consumption, not creation.
```
muXxer: which signature check do you mean?

block.go
```diff
@@ -269,3 +270,19 @@ func (b *Block) DoPOW(ctx context.Context, targetScore float64) error {
...
+	return workScoreStructure.WorkScoreMaxParents + workScoreStructure.FactorMissingParent.Multiply(numMissingParents) +
+		// ProtocolVersion and NetworkID
+		workScoreStructure.FactorData.Multiply(serializer.OneByte+serializer.UInt64ByteSize) +
+		// IssuerID and IssuingTime
+		workScoreStructure.FactorData.Multiply(AccountIDLength+util.NumByteLen(b.IssuingTime)) +
+		// SlotCommitment and LatestFinalizedSlot
+		workScoreStructure.FactorData.Multiply(util.NumByteLen(b.SlotCommitment)+serializer.UInt64ByteSize) +
```
muXxer: `util.NumByteLen(b.SlotCommitment)` would panic. You need to implement a `Size()` for the commitment instead.

block.go
```diff
@@ -269,3 +270,19 @@ func (b *Block) DoPOW(ctx context.Context, targetScore float64) error {
...
 }
+
+func (b *Block) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// Work Score for parents is the score for max parents plus a penalty for any less than max parents
+	numMissingParents := BlockMaxParents - (len(b.StrongParents) + len(b.WeakParents) + len(b.ShallowLikeParents))
+	return workScoreStructure.WorkScoreMaxParents + workScoreStructure.FactorMissingParent.Multiply(numMissingParents) +
+		// ProtocolVersion and NetworkID
+		workScoreStructure.FactorData.Multiply(serializer.OneByte+serializer.UInt64ByteSize) +
+		// IssuerID and IssuingTime
+		workScoreStructure.FactorData.Multiply(AccountIDLength+util.NumByteLen(b.IssuingTime)) +
```
muXxer: `util.NumByteLen(b.IssuingTime)` will panic, use `serializer.UInt64ByteSize` instead.

block.go
```diff
@@ -269,3 +270,19 @@ func (b *Block) DoPOW(ctx context.Context, targetScore float64) error {
...
+		// SlotCommitment and LatestFinalizedSlot
+		workScoreStructure.FactorData.Multiply(util.NumByteLen(b.SlotCommitment)+serializer.UInt64ByteSize) +
+		b.Payload.WorkScore(workScoreStructure) +
+		// BurnedMana and Nonce
+		workScoreStructure.FactorData.Multiply(ManaSize+serializer.UInt64ByteSize) +
+		workScoreStructure.WorkScoreEd25519Signature
```
muXxer: We should add a `WorkScore` function to the signature interface and get the costs from there to be future proof (in case we add a new signature scheme or implement poor man's multisig etc.) This should be used everywhere where you added `WorkScoreEd25519Signature` now.

block.go
```diff
@@ -269,3 +270,19 @@ func (b *Block) DoPOW(ctx context.Context, targetScore float64) error {
...
+	numMissingParents := BlockMaxParents - (len(b.StrongParents) + len(b.WeakParents) + len(b.ShallowLikeParents))
+	return workScoreStructure.WorkScoreMaxParents + workScoreStructure.FactorMissingParent.Multiply(numMissingParents) +
+		// ProtocolVersion and NetworkID
+		workScoreStructure.FactorData.Multiply(serializer.OneByte+serializer.UInt64ByteSize) +
+		// IssuerID and IssuingTime
+		workScoreStructure.FactorData.Multiply(AccountIDLength+util.NumByteLen(b.IssuingTime)) +
+		// SlotCommitment and LatestFinalizedSlot
+		workScoreStructure.FactorData.Multiply(util.NumByteLen(b.SlotCommitment)+serializer.UInt64ByteSize) +
+		b.Payload.WorkScore(workScoreStructure) +
+		// BurnedMana and Nonce
```
muXxer: Nonce can be removed.

feat_blockissuer.go
```diff
@@ -34,6 +34,12 @@ func (s *BlockIssuerFeature) VBytes(rentStruct *RentStructure, _ VBytesFunc) VBy
 		rentStruct.VBFactorKey.Multiply(VBytes(len(s.BlockIssuerKeys))*ed25519.PublicKeySize)
 }
 
+func (s *BlockIssuerFeature) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// Block issuer feature requires invocation of account and mana managers, so requires extra work.
+	return workScoreStructure.FactorData.Multiply(s.Size()) +
```
muXxer: Should we also charge an Offset for all blocks because we need to do mana decay for every block anyway?
And I don't understand if that feature here is accounted for for the input side or output side or both?

feat_issuer.go
```diff
@@ -23,6 +23,10 @@ func (s *IssuerFeature) VBytes(rentStruct *RentStructure, _ VBytesFunc) VBytes {
 		s.Address.VBytes(rentStruct, nil)
 }
 
+func (s *IssuerFeature) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	return workScoreStructure.FactorData.Multiply(s.Size())
```
muXxer: Maybe add a comment that the signature is not needed here, because the signature needs to be unlocked already anyway. This is just a simple lookup.

feat_sender.go
```diff
@@ -23,6 +23,10 @@ func (s *SenderFeature) VBytes(rentStruct *RentStructure, f VBytesFunc) VBytes {
 	return rentStruct.VBFactorData.Multiply(serializer.SmallTypeDenotationByteSize) + s.Address.VBytes(rentStruct, nil)
 }
 
+func (s *SenderFeature) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	return workScoreStructure.FactorData.Multiply(s.Size())
```
muXxer: Maybe add a comment that the signature is not needed here, because the signature needs to be unlocked already anyway. This is just a simple lookup.

feat_staking.go
```diff
@@ -51,6 +51,12 @@ func (s *StakingFeature) VBytes(rentStruct *RentStructure, _ VBytesFunc) VBytes
 	return rentStruct.VBFactorData.Multiply(VBytes(vbytes))
 }
 
+func (s *StakingFeature) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// Staking feature changes require invokation of staking managers so require extra work.
```
muXxer: Can we find a way to not charge workscore costs if the output is simply transitioned without altering the staking feature?

unlock_cond_storage_return.go
```diff
@@ -26,6 +26,11 @@ func (s *StorageDepositReturnUnlockCondition) VBytes(rentStruct *RentStructure,
 		s.ReturnAddress.VBytes(rentStruct, nil)
 }
 
+func (s *StorageDepositReturnUnlockCondition) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// StorageDepositReturnUnlockCondition requires a signature check, but on consumption, not creation.
```
muXxer: But isn't that signature already unlocked anyway? Maybe add the lookup comment here like in the other cases.

unlock_signature.go
```diff
@@ -17,3 +17,7 @@ func (s *SignatureUnlock) Type() UnlockType {
 func (s *SignatureUnlock) Size() int {
 	return util.NumByteLen(byte(UnlockSignature)) + s.Signature.Size()
 }
+
+func (s *SignatureUnlock) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	return workScoreStructure.WorkScoreEd25519Signature
```
muXxer: Add the mentioned Signature.WorkScore() function here.

unlock_nft.go
muXxer: Btw, we never charge for these right? Because we never analyse inputs anyway? Maybe we should switch to charge costs if someone creates these things on the output side.

unlock_reference.go
muXxer: Btw, we never charge for these right? Because we never analyse inputs anyway? Maybe we should switch to charge costs if someone creates these things on the output side.

unlock_signature.go
muXxer: Btw, we never charge for these right? Because we never analyse inputs anyway? Maybe we should switch to charge costs if someone creates these things on the output side.

unlock_cond_storage_return.go
muXxer: Or we switch to a "creator pays principle" and charge everyone that creates an output that has these features.
So we would charge him for the signature check here as well.

unlock_cond_state_ctrl.go
```diff
@@ -19,6 +19,11 @@ func (s *StateControllerAddressUnlockCondition) VBytes(rentStruct *RentStructure
 		s.Address.VBytes(rentStruct, nil)
 }
 
+func (s *StateControllerAddressUnlockCondition) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// StateControllerAddressUnlockCondition requires a signature check, but on consumption, not creation.
+	return workScoreStructure.FactorData.Multiply(s.Size())
```
muXxer: Or we switch to a "creator pays principle" and charge everyone that creates an output that has these features.
So we would charge him for the signature check here as well.

unlock_cond_imm_account.go
```diff
@@ -21,6 +21,11 @@ func (s *ImmutableAccountUnlockCondition) VBytes(rentStruct *RentStructure, _ VB
 		s.Address.VBytes(rentStruct, nil)
 }
 
+func (s *ImmutableAccountUnlockCondition) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// ImmutableAccountUnlockCondition requires a signature check, but on consumption, not creation.
+	return workScoreStructure.FactorData.Multiply(s.Size())
```
muXxer: Or we switch to a "creator pays principle" and charge everyone that creates an output that has these features.
So we would charge him for the signature check here as well.

unlock_cond_governor.go
```diff
@@ -19,6 +19,11 @@ func (s *GovernorAddressUnlockCondition) VBytes(rentStruct *RentStructure, _ VBy
 		s.Address.VBytes(rentStruct, nil)
 }
 
+func (s *GovernorAddressUnlockCondition) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// GovernorAddressUnlockCondition requires a signature to unlock, but this is done when the output is consumed.
+	return workScoreStructure.FactorData.Multiply(s.Size())
```
muXxer: Or we switch to a "creator pays principle" and charge everyone that creates an output that has these features.
So we would charge him for the signature check here as well.

unlock_cond_expiration.go
```diff
@@ -28,6 +28,11 @@ func (s *ExpirationUnlockCondition) VBytes(rentStruct *RentStructure, _ VBytesFu
 		s.ReturnAddress.VBytes(rentStruct, nil)
 }
 
+func (s *ExpirationUnlockCondition) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// Expiration condition requires Commitment inputs to be loaded and signature checks, but this is taken care of when the output is consumes, not created.
+	return workScoreStructure.FactorData.Multiply(s.Size())
```
muXxer: But when do we charge for that?

unlock_cond_address.go
```diff
@@ -19,6 +19,11 @@ func (s *AddressUnlockCondition) VBytes(rentStruct *RentStructure, _ VBytesFunc)
 		s.Address.VBytes(rentStruct, nil)
 }
 
+func (s *AddressUnlockCondition) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// Address require signature check but this is done on consumption of the output, not creation.
+	return workScoreStructure.FactorData.Multiply(s.Size())
```
muXxer: Or we switch to a "creator pays principle" and charge everyone that creates an output that has these features.
So we would charge him for the signature check here as well.

unlock_account.go
```diff
@@ -35,3 +35,7 @@ func (r *AccountUnlock) Type() UnlockType {
 func (r *AccountUnlock) Size() int {
 	return AccountUnlockSize
 }
+
+func (r *AccountUnlock) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	return workScoreStructure.WorkScoreEd25519Signature
```
muXxer: Add the mentioned Signature.WorkScore() function here.

muXxer: Btw, we never charge for these right? Because we never analyse inputs anyway? Maybe we should switch to charge costs if someone creates these things on the output side.

transaction.go
```diff
@@ -157,3 +157,8 @@ func (t *Transaction) syntacticallyValidate(protoParams *ProtocolParameters) err
 
 	return nil
 }
+
+func (t *Transaction) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	return t.Unlocks.WorkScore(workScoreStructure) +
```
muXxer: The FactorData for `util.NumByteLen(uint32(PayloadTransaction))` is missing here.

transaction_essence.go
```diff
@@ -212,3 +213,14 @@ func (u *TransactionEssence) syntacticallyValidate(protoParams *ProtocolParamete
 
 	return nil
 }
+
+func (u *TransactionEssence) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	// NetworkID, CreationTime
```
muXxer: `workScoreStructure.FactorData.Multiply(util.NumByteLen(TransactionEssenceNormal))` is missing here

output_delegation.go
```diff
@@ -147,6 +147,15 @@ func (d *DelegationOutput) VBytes(rentStruct *RentStructure, _ VBytesFunc) VByte
 		d.ImmutableFeatures.VBytes(rentStruct, nil)
 }
 
+func (d *DelegationOutput) WorkScore(workScoreStructure *WorkScoreStructure) WorkScore {
+	return d.Conditions.WorkScore(workScoreStructure) +
+		d.ImmutableFeatures.WorkScore(workScoreStructure) +
+		workScoreStructure.FactorData.Multiply(DelegationIDLength) +
```
muXxer: why not write all in one FactorData.Multiply? That reduces the amount of multiplications.

muXxer (changes requested):

allotment.go
cyberphysic4l: Agreed, done.

block.go
cyberphysic4l: And then don't even take the weak and shallow parents into account at all, even the data factor? I suppose I am fine with this. As long as we don't even include the data factor, this doesn't discourage larger number of parents.

workscore.go
cyberphysic4l: good suggestion, done.

api.go
cyberphysic4l: indeed.

unlock_signature.go
cyberphysic4l: We do charge for these in when calculating the workscore on the transaction essence.

unlock_account.go
cyberphysic4l: I think this one shoudl not actually incur a signature cost because it will already be accounted for by the signature unlock of the account output.

cyberphysic4l: Does that sound correct? Same for other reference unlocks...

feat_staking.go
cyberphysic4l: I don't think this is possible without resolving the inputs.

unlock_cond_storage_return.go
cyberphysic4l: Updated the comment. We will only charge for consumption by referring to the unlocks rather than unlock conditions.

api.go
cyberphysic4l: done.

workscore.go
cyberphysic4l: Which part are you referring to removing here?

unlock_nft.go
cyberphysic4l: I changed this now to just be a data factor for work score rather than charging for signature.

unlock_reference.go
cyberphysic4l: changed now to just be data.

block.go
cyberphysic4l: Done.

feat_blockissuer.go
cyberphysic4l: The feature is accounted for only on the output side, so similar to staking feature, we charge for touching an output with block issuer feature even if the values aren't changed. I don't think there is any way we can charge for if depending on the transition...

cyberphysic4l: An offset for blocks is probably a good idea...

feat_issuer.go
cyberphysic4l: Added a comment.

unlock_signature.go
cyberphysic4l: Added the Signature.WorkScore() func

unlock_cond_state_ctrl.go
cyberphysic4l: I think we now agreed we will go with consumer pays and charge for signatures through the unlocks.

unlock_cond_imm_account.go
cyberphysic4l: I think we now agreed we will go with consumer pays and charge for signatures through the unlocks.

unlock_cond_governor.go
cyberphysic4l: I think we now agreed we will go with consumer pays and charge for signatures through the unlocks.

unlock_cond_address.go
cyberphysic4l: I think we now agreed we will go with consumer pays and charge for signatures through the unlocks.

transaction.go
cyberphysic4l: Added

output_delegation.go
cyberphysic4l: done

unlock_cond_expiration.go
cyberphysic4l: This is taken into account by the cost of the unlocks and the cost of the ContextInputs provided to unlock this output.

workscore.go
muXxer: The `WorkScoreMaxParents`, but it is already gone now.

unlock_account.go
muXxer: correct

feat_blockissuer.go
muXxer: We will account for the mana decay in the Inputs factor and we add another WorkScore for Block.

feat_staking.go
muXxer: Maybe we can find a solution with future upgrades if that turns out to be a problem.

muXxer (approved):

PR #454, opened 2023-07-12
Separate input and context input types

PhilippGackstatter:
# Description of change

Separate inputs from context inputs more clearly in the type system. This PR makes `UTXOInput` implement the `Input` interface, while `RewardInput`, `CommitmentInput` and `BlockIssuanceCreditInput` (renamed from `BICInput`) implement the `ContextInput` interface. There is however no interface abstracting over all input types.
The motivation for this change is that inputs and context inputs are separate fields in the transaction but the type system currently does not differentiate between the two input types.
Additionally, so far, inputs and context inputs shared the same type range (the prefixed byte identifying the type), which is unnecessary given that they are serialized in two entirely different fields in the transaction.

## Type of change

Adapted and wrote tests.

## Change checklist


input.go
```diff
@@ -29,15 +23,15 @@ func (inputType InputType) String() string {
 }
 
 var (
-	inputNames = [InputBlockIssuanceCredit + 1]string{"UTXOInput", "CommitmentInput", "BICInput"}
+	inputNames = [InputUTXO + 1]string{"UTXOInput"}
```
piotrm50, line 26: Do we even need this part `[InputUTXO + 1]`?

piotrm50 (approved):

input.go
PhilippGackstatter: It also works with just `[]` but I think it doesn't hurt to have it. If anything, it makes it a bit clearer? Don't have a strong opinion on it though.

PR #456, opened 2023-07-14
Align Mana field in Outputs

PhilippGackstatter:
# Description of change

Align Mana field with TIPs.

- Change the field's position in Basic, Account and NFT Outputs.
- Remove Mana field from Foundry.

## Type of change

`go test`

## Change checklist

### DIFF
--- a/output_foundry.go
+++ b/output_foundry.go
@@ -95,8 +95,6 @@ type FoundryOutput struct {
 	Features FoundryOutputFeatures `serix:"5,mapKey=features,omitempty"`
 	// The immutable feature on the output.
 	ImmutableFeatures FoundryOutputImmFeatures `serix:"6,mapKey=immutableFeatures,omitempty"`
-	// The stored mana held by the output.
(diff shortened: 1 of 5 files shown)


output_foundry.go
```diff
@@ -200,7 +197,7 @@ func (f *FoundryOutput) Deposit() BaseToken {
 }
 
 func (f *FoundryOutput) StoredMana() Mana {
```
daria305, line 199: Isn't this strange we have "mana" in general interface if not all outputs has mana. I believe removing `StoredMana` from `Output` interface could be problematic in implementation, but at least does not imply all outputs has the ability to store mana.

daria305 (commented):
Seams good, foundry out never was supposed to keep mana, I added just one discussion point

output_foundry.go
PhilippGackstatter: It's a bit strange, but still fine. Delegation and Foundry Outputs don't hold Mana, but since this is only a getter, it can just return 0.

daria305 (approved):

PR #463, opened 2023-07-26
Remove immutable features on `DelegationOutput`

PhilippGackstatter:
# Description of change

Removes the immutable features on `DelegationOutput`.

The `IssuerFeature` was the only allowed feature in the immutable feature set. However, it is not useful to have since a delegation output is usually not transferred, where the issuer feature would have its use.

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

Adapted tests.

## Change checklist


output_delegation.go
```diff
@@ -164,7 +158,7 @@ func (d *DelegationOutput) UnlockConditionSet() UnlockConditionSet {
 }
 
 func (d *DelegationOutput) ImmutableFeatureSet() FeatureSet {
```
muXxer: In my opinion this function should be removed from the delegation output.
You should create a new interface for chained outputs with the immutable feature.

```diff

// ChainOutput is a type of Output which represents a chain of state transitions.
type ChainOutput interface {
	Output
	// Chain returns the ChainID to which this Output belongs to.
	Chain() ChainID
-	// ImmutableFeatureSet returns the immutable FeatureSet this output contains.
-	ImmutableFeatureSet() FeatureSet
}

+// ChainOutputImmutable is a type of Output which represents a chain of state transitions with immutable features.
+type ChainOutputImmutable interface {
+	ChainOutput
+	// ImmutableFeatureSet returns the immutable FeatureSet this output contains.
+	ImmutableFeatureSet() FeatureSet
+}
```

Then you can remove the `return vm.IsIssuerOnOutputUnlocked(current, vmParams.WorkingSet.UnlockedIdents)` check in `delegationGenesisValid`.

output_delegation.go
```diff
@@ -143,8 +138,7 @@ func (d *DelegationOutput) VBytes(rentStruct *RentStructure, _ VBytesFunc) VByte
 		rentStruct.VBFactorData.Multiply(serializer.SmallTypeDenotationByteSize+serializer.UInt64ByteSize*4) +
```
muXxer, line 137: Mhh your current commit doesn't contain the workscore changes. You should pull in current `serix-iota-core` and remove the immutable features there as well.

muXxer (changes requested):

muXxer (approved):

PR #468, opened 2023-08-01
Use serix to serialize JSON apimodels

### LINKED ISSUE #230: Node client API version initialization

alexsporn:


go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/iotaledger/hive.go/lo v0.0.0-20230728082521-74f3f2957f0c
 	github.com/iotaledger/hive.go/runtime v0.0.0-20230728082521-74f3f2957f0c
 	github.com/iotaledger/hive.go/serializer/v2 v2.0.0-rc.1.0.20230728100039-3d97aada979b
+	github.com/iotaledger/hive.go/stringify v0.0.0-20230728082147-9fbb11f3eab5
```
muXxer: Better use the same hive.go version as the other modules `20230728082521-74f3f2957f0c`
Not sure if we fixed the exp madness there.

api/epoch_based_provider.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (e *EpochBasedProvider) VersionForEpoch(epoch iotago.EpochIndex) iotago.Version {
+	e.mutex.RLock()
+	defer e.mutex.RUnlock()
+
+	return e.protocolVersions.VersionForEpoch(epoch)
+}
+
+func WithAPIForMissingVersionCallback(callback func(version iotago.Version) (iotago.API, error)) options.Option[EpochBasedProvider] {
```
muXxer: I like when the possible options are at the top, before the constructor of the object that takes the options :sweat_smile:

api/epoch_based_provider.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (e *EpochBasedProvider) AddProtocolParametersAtEpoch(protocolParameters iotago.ProtocolParameters, epoch iotago.EpochIndex) {
+	e.AddProtocolParameters(protocolParameters)
+	e.AddVersion(protocolParameters.Version(), epoch)
+}
+
+func (e *EpochBasedProvider) AddProtocolParameters(protocolParameters iotago.ProtocolParameters) {
+	e.mutex.Lock()
+	defer e.mutex.Unlock()
```
muXxer: I think that can actually deadlock, because if someone tries to read the Version in parallel, the latestVersionMutex might be readlocked, but can't be unlocked because it waits for the mutex readlock to be aquired, which is locked here.
So better don't use defer, but unlock after the delete.

muXxer (changes requested):

nodeclient/apimodels/core_test.go
```diff
@@ -0,0 +1,352 @@
...
+			Name:            "IOTA",
+			TickerSymbol:    "IOTA",
+			Unit:            "MIOTA",
+			UseMetricPrefix: true,
+		}
+
+		jsonResponse, err := api.JSONEncode(response)
+		require.NoError(t, err)
+
+		expected := "{\"name\":\"IOTA\",\"tickerSymbol\":\"IOTA\",\"unit\":\"MIOTA\",\"decimals\":0,\"useMetricPrefix\":true}"
```
muXxer, line 85: maybe add a comment that subunit is empty.... I was wondering how that test works :D

nodeclient/apimodels/indexer.go
```diff
@@ -9,13 +9,13 @@ import (
...
 type IndexerResponse struct {
 	// The ledger index at which these outputs where available at.
-	LedgerIndex iotago.SlotIndex `json:"ledgerIndex"`
+	LedgerIndex iotago.SlotIndex `serix:"0,mapKey=ledgerIndex"`
 	// The maximum count of results that are returned by the node.
-	PageSize int `json:"pageSize"`
+	PageSize uint32 `serix:"1,mapKey=pageSize"`
 	// The output IDs (transaction hash + output index) of the found outputs.
-	Items iotago.HexOutputIDs `json:"items"`
+	Items iotago.HexOutputIDs `serix:"2,mapKey=items,lengthPrefixType=uint8"`
```
muXxer: isn't 255 max results a bit too small?

muXxer (changes requested):

muXxer (approved):

alexsporn:
Fixed https://github.com/iotaledger/iota-core/issues/230

PR #470, opened 2023-08-01
Adapt serix to api responses

jkrvivian:
# Description of change

uint64 and int64 are serialized to string with serix, for consistency, we'll use serix instead of `encoding/json` in responses  as well.
Relevant API implementation and spec updates would be made in iota-core and TIPs.

## Type of change
- Enhancement


jkrvivian:
It's already part of #468

PR #479, opened 2023-08-10
Add scheduler params and max block work and activate workscore

cyberphysic4l:
companion PR for https://github.com/iotaledger/iota-core/pull/283


cyberphysic4l:
Now that Block interface is a Sizer, there is a failing test related to deserialisation: the serialiser yields a different number of bytes that the Size() method for blocks. Off by 7 for basic blocks and off by 3 for validation blocks.

api_v3_protocol_parameters.go
```diff
@@ -183,7 +183,7 @@ func WithWorkScoreOptions(
 	minStrongParentsThreshold byte) options.Option[V3ProtocolParameters] {
 	return func(p *V3ProtocolParameters) {
 		p.basicProtocolParameters.WorkScoreStructure = WorkScoreStructure{
-			DataByte:                  dataByte,
+			DataKilobyte:              dataKilobyte,
```
piotrm50, line 185: Why kilobyte?

block.go
```diff
@@ -127,14 +127,12 @@ func (b *BlockHeader) Hash(api API) (Identifier, error) {
...
-func (b *BlockHeader) WorkScore(workScoreStructure *WorkScoreStructure) (WorkScore, error) {
-	// Version, NetworkID, IssuingTime, SlotCommitmentID, LatestFinalisedSlot and IssuerID
-	return workScoreStructure.DataByte.Multiply(serializer.OneByte +
-		serializer.UInt64ByteSize +
-		serializer.UInt64ByteSize +
-		CommitmentIDLength +
-		SlotIndexLength +
-		AccountIDLength)
+func (b *BlockHeader) WorkScore(_ *WorkScoreStructure) (WorkScore, error) {
+	return 0, nil
```
piotrm50, line 131: Is block header always free?

block.go
```diff
@@ -279,6 +277,12 @@ func (b *ProtocolBlock) ForEachParent(consumer func(parent Parent)) {
 }
 
 func (b *ProtocolBlock) WorkScore(workScoreStructure *WorkScoreStructure) (WorkScore, error) {
+	workScoreBytes, err := workScoreStructure.DataKilobyte.Multiply(b.Size())
```
piotrm50, line 280: WorkScore seems to be calculated twice - once here because of `b.Size()` which includes Header, Signature and the internal Block and again below when we retrieve WorkScore separately. That seems incorrect.

feat.go
```diff
@@ -85,25 +85,20 @@ func (f Features[T]) VBytes(rentStruct *RentStructure, _ VBytesFunc) VBytes {
 }
 
 func (f Features[T]) WorkScore(workScoreStructure *WorkScoreStructure) (WorkScore, error) {
-	// LengthPrefixType
-	workScoreBytes, err := workScoreStructure.DataByte.Multiply(serializer.OneByte)
-	if err != nil {
-		return 0, err
-	}
-
+	var workScoreFeats WorkScore
```
piotrm50, line 88: Why is LengthPrefixType not included here any more? Is it included by each feature's `WorkScore` method?

feat_blockissuer.go
```diff
@@ -43,13 +43,8 @@ func (s *BlockIssuerFeature) VBytes(rentStruct *RentStructure, _ VBytesFunc) VBy
...
 
 func (s *BlockIssuerFeature) WorkScore(workScoreStructure *WorkScoreStructure) (WorkScore, error) {
-	workScoreBytes, err := workScoreStructure.DataByte.Multiply(s.Size())
-	if err != nil {
-		return 0, err
-	}
-
 	// block issuer feature requires invocation of account and mana managers, so requires extra work.
-	return workScoreBytes.Add(workScoreStructure.BlockIssuer)
+	return workScoreStructure.BlockIssuer, nil
```
piotrm50, line 47: Doesn't it depend on the size of the feature? Like number of keys etc?

piotrm50 (approved):

api_v3_protocol_parameters.go
cyberphysic4l: This allows us to assign workscore less than 1 per byte.

block.go
cyberphysic4l: The BlockHeader's data will still be charged for in the highest level object, the protocol block, where the entire size of the protocol block, including block header within, is used to get the data factor of the workscore.

block.go
cyberphysic4l: We only apply the data factor once for the highest level object, the protocol block, because this allows us to avoid losing precision when we divide by 1024 for the kilobyte factor.

feat.go
cyberphysic4l: This only applies to data factor, which is now applied only to the protocol block rather than all subcomponents. LengthPrefixType is included in the Size() of the features, so it is accounted for.

feat_blockissuer.go
cyberphysic4l: same as above

PR #483, opened 2023-08-16
Let Governor control the Block Issuer Feature

PhilippGackstatter:
# Description of change

Change permissions for governor and state controller with respect to the Block Issuer Feature (BIF) and Staking Feature (SF).

- The BIF can be modified by the governor.
- The SF can be modified by state controller.

Additionally, handle other cases:
- Check for negative BIC (= account locked) more generally, even when BIF is untouched in the transaction. Introduce dedicated error types for these cases making it easier to test for them and those tests more precise.
- Staking & block issuer feature creation during regular account transitions. Previously this was only handled during account genesis itself.

## Type of change

Added various STVF tests and adapted existing ones.

## Change checklist

### DIFF
--- a/vm/stardust/vm_stardust.go
+++ b/vm/stardust/vm_stardust.go
@@ -203,11 +203,7 @@ func accountGovernanceSTVF(input *vm.ChainOutputWithCreationSlot, next *iotago.A
 		return ierrors.Wrapf(iotago.ErrInvalidAccountGovernanceTransition, "%w", iotago.ErrInvalidStakingBlockIssuerRequired)
 	}
 
-	if err := accountBlockIssuerSTVF(input, next, vmParams); err != nil {
-		return err
-	}
-
-	return nil
+	return accountBlockIssuerSTVF(input, next, vmParams)
 }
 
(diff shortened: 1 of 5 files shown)


vm/stardust/vm_stardust.go
```diff
@@ -201,8 +224,8 @@ func accountStateSTVF(input *vm.ChainOutputWithCreationSlot, next *iotago.Accoun
 		return ierrors.Wrapf(iotago.ErrInvalidAccountStateTransition, "%w", err)
 	}
 
-	if err := accountBlockIssuerSTVF(input, next, vmParams); err != nil {
-		return err
+	if err := iotago.FeatureUnchanged(iotago.FeatureBlockIssuer, current.Features.MustSet(), next.Features.MustSet()); err != nil {
+		return ierrors.Wrapf(iotago.ErrInvalidAccountStateTransition, "%w", err)
```
cyberphysic4l, line 230: It would be nice to put a bit of explanation as to why the block issuer feature is not allowed to change here. For example, "block issuer feature cannot change during account state transition", or "block issuer feature can only be changed with governance transition"

cyberphysic4l (approved):

PR #485, opened 2023-08-16
Require lexical order for `Allotments`

PhilippGackstatter:
# Description of change

Require lexical order for `Allotments`. The argument in favor of lexical order is to more easily check whether two transactions are the same. This is also required for many other arrays, so this is a move towards more consistency.

This is implemented using the hive serializer and the custom uniqueness check for allotments is removed. An interesting finding is also that the behavior of the serializer's lexical order check is different when used with no duplicates or without. With no duplicates mode, the lexical order is checked using the same bytes as uniqueness is checked for (based on `UniquenessSliceFunc`), while all bytes are used when duplicates mode is not activated. Fortunately, this is exactly what we need for allotments, where the Account ID is what matters for uniqueness and lexical order, analogous to native tokens.

This PR also fixes an error in the (de)serialize test implementation, where the serialization and deserialization errors were not actually correctly checked (`require.Error` was used instead of `require.ErrorIs`). Fortunately, no test was actually broken.

## Type of change

Existing tests pass or were adapted.

## Change checklist

### DIFF
--- a/allotment.go
+++ b/allotment.go
@@ -1,7 +1,9 @@
 package iotago
 
 import (
-	"github.com/iotaledger/hive.go/ierrors"
+	"bytes"
+	"sort"
+
 	"github.com/iotaledger/hive.go/serializer/v2"
 )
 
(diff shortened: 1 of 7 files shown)


piotrm50 (approved):
lgtm

PR #487, opened 2023-08-22
Change TimeProvider's perception of time around Genesis, slot 0-1 and epoch 0-1

jonastheis:
This PR fixes the `TimeProvider` and its perception of time, slots and epochs and how they relate to each other. 
Before, there were some inconsistencies especially around slot 0 and 1 and epoch 0 and 1. 

Now a new way for handling time is introduced.
```
Slots are counted starting from 1 with 0 being reserved for times before the genesis, which has to be addressable as its own slot.
Epochs are counted starting from 0.

Example: with slotDurationSeconds = 10 and slotsPerEpochExponent = 3
[] inclusive range boundary, () exclusive range boundary
slot 0: [-inf; genesis)
slot 1: [genesis; genesis+10)
slot 2: [genesis+10; genesis+20)
...
epoch 0: [slot 0; slot 8)
epoch 1: [slot 8; slot 16)
epoch 2: [slot 16; slot 24)
...
```

### DIFF
--- a/timeprovider.go
+++ b/timeprovider.go
@@ -27,7 +27,7 @@ type TimeProvider struct {
 func NewTimeProvider(genesisUnixTime int64, slotDurationSeconds int64, slotsPerEpochExponent uint8) *TimeProvider {
 	// if slotDurationSeconds == 0 {
 	//	panic("slot duration can't be zero")
-	//}
+	// }
 
 	return &TimeProvider{
 		genesisUnixTime:       genesisUnixTime,
@@ -98,6 +98,10 @@ func (t *TimeProvider) SlotEndTime(i SlotIndex) time.Time {
 
 // EpochFromSlot calculates the EpochIndex from the given slot.
 func (t *TimeProvider) EpochFromSlot(slot SlotIndex) EpochIndex {
+	if slot == 0 {
+		return 0
+	}
+
 	return EpochIndex(slot>>SlotIndex(t.slotsPerEpochExponent) + 1)
 }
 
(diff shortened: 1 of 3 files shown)


jkrvivian (approved):

timeprovider.go
```diff
@@ -4,11 +4,27 @@ import (
...
+// Example: with slotDurationSeconds = 10 and slotsPerEpochExponent = 3
+// <> inclusive range boundary, () exclusive range boundary
+// slot 0: <-inf; genesis)
+// slot 1: <genesis; genesis+10)
+// slot 2: <genesis+10; genesis+20)
+// ...
+// epoch 0: <slot 0; slot 8)
+// epoch 1: <slot 8; slot 16)
+// epoch 2: <slot 16; slot 24)
+// ...
```
PhilippGackstatter:
```suggestion
// Example: with slotDurationSeconds = 10 and slotsPerEpochExponent = 3
// [] inclusive range boundary, () exclusive range boundary
// slot 0: [-inf; genesis)
// slot 1: [genesis; genesis+10)
// slot 2: [genesis+10; genesis+20)
// ...
// epoch 0: [slot 0; slot 8)
// epoch 1: [slot 8; slot 16)
// epoch 2: [slot 16; slot 24)
// ...
```

Nit: The conventional way to write inclusive and exclusive ranges is with `[]` and `()` respectively. Would be a bit nicer to use that here.

timeprovider.go
```diff
@@ -27,10 +43,11 @@ type TimeProvider struct {
 func NewTimeProvider(genesisUnixTime int64, slotDurationSeconds int64, slotsPerEpochExponent uint8) *TimeProvider {
 	// if slotDurationSeconds == 0 {
 	//	panic("slot duration can't be zero")
-	//}
+	// }
```
PhilippGackstatter: Can we get rid of this?

PhilippGackstatter (approved):

piotrm50 (approved):

PR #489, opened 2023-08-29
Block Issuer Keys interface

PhilippGackstatter:
# Description of change

Add a Block Issuer Keys interface that is implemented by Ed25519 keys. This is so we can add a type prefix to the serialized public key and are more future proof.

## Type of change

Existing tests run successfully. Manually checked the produced JSON has the expected type prefix.

## Change checklist


vm/stardust/stvf_test.go
```diff
@@ -1970,7 +1970,7 @@ func TestAccountOutput_ValidateStateTransition(t *testing.T) {
 				Features: iotago.AccountOutputFeatures{
 					&iotago.SenderFeature{Address: exampleStateCtrl},
 					&iotago.BlockIssuerFeature{
-						BlockIssuerKeys: []ed25519.PublicKey{tpkg.Rand32ByteArray()},
+						BlockIssuerKeys: []iotago.BlockIssuerKey{iotago.BlockIssuerKeyEd25519FromPublicKey(tpkg.Rand32ByteArray())},
```
muXxer: You could add a `RandomBlockIsssuerKeysEd25519(count)` to simplify the tests.

block_issuer_key.go
```diff
@@ -0,0 +1,51 @@
...
+	for _, key := range keys {
+		size += key.Size()
+	}
+
+	return size
+}
+
+func (keys BlockIssuerKeys) VBytes(rentStruct *RentStructure, _ VBytesFunc) VBytes {
+	// VBFactorIssuerKeys: numKeys * (type prefix + pubKeyLength)
+	return rentStruct.VBFactorIssuerKeys.Multiply(VBytes(len(keys)) * (serializer.TypeDenotationByteSize + ed25519.PublicKeySize))
```
muXxer: ed25519 PublicKeySize? Thats the generic type here...

block_issuer_key.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (keys BlockIssuerKeys) Sort() {
+	sort.Slice(keys, func(i, j int) bool {
+		return bytes.Compare(keys[i].PublicKeyBytes(), keys[j].PublicKeyBytes()) < 0
+	})
+}
+
+// Size returns the size of the public key when serialized.
+func (keys BlockIssuerKeys) Size() int {
+	size := 0
```
muXxer: Isn't the length prefix missing here?

block_issuer_key.go
muXxer: Isn't the length prefix missing here?

block_issuer_key.go
```diff
@@ -0,0 +1,51 @@
...
+	// Ed25519BlockIssuerKey denotes a BlockIssuerKeyEd25519.
+	Ed25519BlockIssuerKey BlockIssuerKeyType = iota
+)
+
+// BlockIssuerKeys are the keys allowed to issue blocks from an account with a BlockIssuerFeature.
+type BlockIssuerKeys []BlockIssuerKey
+
+func (keys BlockIssuerKeys) Sort() {
+	sort.Slice(keys, func(i, j int) bool {
+		return bytes.Compare(keys[i].PublicKeyBytes(), keys[j].PublicKeyBytes()) < 0
```
muXxer: Should we maybe compare the Bytes instead of the PublicKeyBytes? I mean the type byte added as well.
What if there are two different types, but they use the same bytes for the pubkey? Then sorting is not guranteed.

block_issuer_key_ed25519.go
```diff
@@ -0,0 +1,36 @@
...
+	return key.PublicKey[:]
+}
+
+// Type returns the BlockIssuerKeyType.
+func (key BlockIssuerKeyEd25519) Type() BlockIssuerKeyType {
+	return Ed25519BlockIssuerKey
+}
+
+// Size returns the size of the public key when serialized.
+func (key BlockIssuerKeyEd25519) Size() int {
```
muXxer, line 36: You should add the VBytes here and use that in the `block_issuer_key.go`

feat_blockissuer.go
```diff
@@ -37,9 +26,9 @@ func (s *BlockIssuerFeature) Clone() Feature {
 
 func (s *BlockIssuerFeature) VBytes(rentStruct *RentStructure, _ VBytesFunc) VBytes {
 	// VBFactorData: type prefix + expiry slot + keys length
-	// VBFactorIssuerKeys: numKeys * pubKeyLength
+	// + block issuer keys
 	return rentStruct.VBFactorData.Multiply(serializer.SmallTypeDenotationByteSize+serializer.UInt64ByteSize+serializer.OneByte) +
```
muXxer: I think the keys length is part of the `BlockIssuerKeys.VBytes` now.

feat_blockissuer.go
```diff
@@ -56,7 +45,7 @@ func (s *BlockIssuerFeature) Equal(other Feature) bool {
 		return false
 	}
 	for i := range s.BlockIssuerKeys {
-		if !bytes.Equal(s.BlockIssuerKeys[i][:], otherFeat.BlockIssuerKeys[i][:]) {
+		if !bytes.Equal(s.BlockIssuerKeys[i].PublicKeyBytes(), s.BlockIssuerKeys[i].PublicKeyBytes()) {
```
muXxer: Better compare the bytes with type instead!

feat_blockissuer.go
```diff
@@ -69,6 +58,6 @@ func (s *BlockIssuerFeature) Type() FeatureType {
 }
 
 func (s *BlockIssuerFeature) Size() int {
-	// FeatureType + BlockIssuerKeys + ExpirySlot
-	return serializer.SmallTypeDenotationByteSize + serializer.OneByte + len(s.BlockIssuerKeys)*ed25519.PublicKeySize + SlotIndexLength
+	// FeatureType + BlockIssuerKeysLengthPrefix + BlockIssuerKeys + ExpirySlot
```
muXxer: I think the keys length is part of the `BlockIssuerKeys.Size` now.

muXxer (changes requested):

muXxer (approved):

PR #493, opened 2023-09-04
Mana & Rewards protocol parameters

karimodm:


mana.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	result, err = safemath.SafeMul(result, uint64(generationRate))
+	if err != nil {
+		return 0, ierrors.Wrap(err, "failed to calculate target reward due to multiplication with generationRate overflow")
+	}
+	subExponent, err := safemath.SafeSub(generationRateExponent, slotsPerEpochExponent)
+	if err != nil {
+		return 0, ierrors.Wrap(err, "failed to calculate target reward due to generationRateExponent - slotsPerEpochExponent subtraction overflow")
+	}
+	finalReward := result >> subExponent
```
oliviasaa: all these lines result in a value of `finalReward` that is constant each time you call this function. Has anyone checked if there is a better place to move them? Can this be a parameter?

karimodm: Moved to a one-time initialization inside `WithRewardsOptions`

mana.go
```diff
@@ -0,0 +1,73 @@
...
+	// RewardsManaShareCoefficient is the coefficient used for calculation of initial rewards, relative to the term theta/(1-theta) from the Whitepaper, with theta = 2/3.
+	RewardsManaShareCoefficient uint64 `serix:"3,mapKey=rewardsManaShareCoefficient"`
+	// DecayBalancingConstantExponent is the exponent used for calculation of the initial reward.
+	DecayBalancingConstantExponent uint8 `serix:"4,mapKey=decayBalancingConstantExponent"`
+	// DecayBalancingConstant needs to be an integer approc  calculated based on chosen DecayBalancingConstantExponent.
+	DecayBalancingConstant uint64 `serix:"5,mapKey=decayBalancingConstant"`
+	// PoolCoefficientExponent is the exponent used for shifting operation in the pool rewards calculations.
+	PoolCoefficientExponent uint8 `serix:"6,mapKey=poolCoefficientExponent"`
+	// ComputedInitialReward is the initial reward calculated from the parameters.
+	ComputedInitialReward uint64
```
alexsporn: We should move computed values out of the serialized params and into the API object holding the params.
Else you will have problems when loading serialized protocol parameters since you won't have the computed values and will need to mutate the params again to update those.

api.go
```diff
@@ -104,6 +108,8 @@ type ProtocolParameters interface {
 	NetworkName() string
 	// NetworkID defines the ID of the network which is derived from the network name.
 	NetworkID() NetworkID
+	// SlotsPerEpochExponent defines the exponent used to calculate the number of slots per epoch.
+	SlotsPerEpochExponent() uint8
```
alexsporn: This seems lost here. Shouldn't it be in the TimeProvider?

alexsporn: if we really need this exposed, then please move it below to the level of the `TimeProvider()`

alexsporn (approved):

PR #497, opened 2023-09-06
Added new RestrictedEd25519Address and ImplicitAccountCreationAddress

### LINKED ISSUE #278: Implicit account creation mechanism
https://hackmd.io/@IF/Sy1ZPiJ2n

alexsporn:


address_implicit_account_creation.go
```diff
@@ -0,0 +1,141 @@
...
+type ImplicitAccountCreationAddress Ed25519Address
+
+// ParseImplicitAccountCreationAddressFromHexString parses the given hex string into an ImplicitAccountCreationAddress.
+func ParseImplicitAccountCreationAddressFromHexString(hexAddr string) (*ImplicitAccountCreationAddress, error) {
+	addrBytes, err := hexutil.DecodeHex(hexAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(addrBytes) < Ed25519AddressBytesLength {
```
PhilippGackstatter: Can we introduce a type alias for these two and use them here instead?

```go
type ImplicitAccountCreationAddressBytesLength = Ed25519AddressBytesLength
type ImplicitAccountCreationAddressSerializedBytesSize = Ed25519AddressSerializedBytesSize
```

address_restricted_ed25519.go
```diff
@@ -0,0 +1,241 @@
...
+	canReceiveOutputsWithExpirationUnlockConditionBitIndex
+	canReceiveOutputsWithStorageDepositReturnUnlockConditionBitIndex
+	canReceiveAccountOutputsBitIndex
+	canReceiveNFTOutputsBitIndex
+	canReceiveDelegationOutputsBitIndex
+)
+
+type RestrictedEd25519Address struct {
+	PubKeyHash   [Ed25519AddressBytesLength]byte `serix:"0"`
+	Capabilities []byte                          `serix:"1,lengthPrefixType=uint8,maxLen=1"`
```
PhilippGackstatter:
```suggestion
	AllowedCapabilities []byte                          `serix:"1,lengthPrefixType=uint8,maxLen=1"`
```

I think calling it `AllowedCapabilites` would be clearer to indicate that it's an AllowList rather than a DenyList.

address_restricted_ed25519.go
```diff
@@ -0,0 +1,241 @@
...
+func (redAddr *RestrictedEd25519Address) CanReceiveNFTOutputs() bool {
+	return redAddr.hasBit(canReceiveNFTOutputsBitIndex)
+}
+
+func (redAddr *RestrictedEd25519Address) CanReceiveDelegationOutputs() bool {
+	return redAddr.hasBit(canReceiveDelegationOutputsBitIndex)
+}
+
+// RestrictedEd25519AddressFromPubKey returns the address belonging to the given Ed25519 public key.
+func RestrictedEd25519AddressFromPubKey(pubKey ed25519.PublicKey,
```
PhilippGackstatter: I think we should have another constructor from a pub key that does not take flags and sets them all to `0` by default.

address_capabilities_bitmask.go
```diff
@@ -0,0 +1,73 @@
...
+	byteIndex := bit / 8
+	if len(bm) <= byteIndex {
+		return false
+	}
+	bitIndex := bit % 8
+	return bm[byteIndex]&(1<<bitIndex) > 0
+}
+
+func (bm AddressCapabilitiesBitMask) setBit(bit int) AddressCapabilitiesBitMask {
+	newBitmask := bm
```
muXxer, line 75: should we maybe check here that `bit` isn't negative?

address_restricted_ed25519.go
```diff
@@ -0,0 +1,157 @@
...
+	canReceiveOutputsWithStorageDepositReturnUnlockCondition bool,
+	canReceiveAccountOutputs bool,
+	canReceiveNFTOutputs bool,
+	canReceiveDelegationOutputs bool) *RestrictedEd25519Address {
+
+	address := blake2b.Sum256(pubKey[:])
+	addr := &RestrictedEd25519Address{}
+	copy(addr.PubKeyHash[:], address[:])
+
+	if canReceiveNativeTokens {
```
muXxer: Should me make an utility function for the following part?
Pass the bools and return the capabilities slice?

address_capabilities_bitmask.go
alexsporn: let me change it to uint

muXxer (approved):

alexsporn:
This implements part of https://github.com/iotaledger/iota-core/issues/278

PR #502, opened 2023-09-13
Add new address type called MultiAddress (trivial multisig, 2FA, etc...)

muXxer:
This PR adds a new address type called `MultiAddress` (also the restricted version of it).

With a `MutliAddress`, users can specify a multi address unlock condition with `k out of n`, where `k` is the threshold and `n` is the cumulative weight of all addresses. If the corresponding `MultiUnlock` inside a transaction contains enough valid unlocks of which the cumulative weight passes the threshold, the `MultiAddress` is valid and the output is unlocked. This enables use cases like trivial multisig, trivial threshold sigs, 2FA, accountability for DID, etc.

By choosing the correct weight for the address, it is possible to create mandatory addresses and optional addresses. 
The limit for different addresses per multi address is set to 10. 

Possible addresses inside a `MultiAddress` are Ed25519, NFT or Account Addresses. It is not allowed to nest a `MultiAddress` inside a `MultiAddress`.

Update: The PR removes the restricted NFT and Account address types, and merges that logic into the `MultiAddress`. This way we keep the protocol simple, and the `restriction` functionality can be achieved for every address type by creating a restricted `MultiAddress` with the wanted address type inside the `MultiAddress` container. The `RestrictedEd25519Address` was kept, because the use case is still valid for e.g. exchanges.

Update: Signature uniqueness checks in the unlocks are now done via the public key in the unlock block, instead of the whole serialized signature block.


address_test.go
```diff
@@ -74,22 +79,56 @@ func TestAddressDeSerialize(t *testing.T) {
...
 	},
 	{
 		"RFC example: Ed25519 testnet",
 		iotago.PrefixDevnet,
 		&iotago.Ed25519Address{0x52, 0xfd, 0xfc, 0x07, 0x21, 0x82, 0x65, 0x4f, 0x16, 0x3f, 0x5f, 0x0f, 0x9a, 0x62, 0x1d, 0x72, 0x95, 0x66, 0xc7, 0x4d, 0x10, 0x03, 0x7c, 0x4d, 0x7b, 0xbb, 0x04, 0x07, 0xd1, 0xe2, 0xc6, 0x49},
 		"atoi1qpf0mlq8yxpx2nck8a0slxnzr4ef2ek8f5gqxlzd0wasgp73utryjjl77h3",
+		nil,
+	},
+	{
+		"Mutli Address",
```
alexsporn: typo

unlock.go
```diff
@@ -143,24 +165,91 @@ func UnlocksSigUniqueAndRefValidator(api API) UnlockValidatorFunc {
...
+						return ierrors.Wrapf(ErrSigUnlockHasNilSig, "at index %d.%d is nil", index, subIndex)
+					}
+
+					sigBlockBytes, err := api.Encode(subUnlock.Signature)
+					if err != nil {
+						return ierrors.Errorf("unable to serialize signature unlock block at index %d.%d for dup check: %w", index, subIndex, err)
+					}
+
+					if existingIndex, exists := seenSigUnlockBytes[string(sigBlockBytes)]; exists {
+						return ierrors.Wrapf(ErrSigUnlockNotUnique, "signature unlock block at index %d.%d is the same as %d", index, subIndex, existingIndex)
```
alexsporn, line 232: This is a problem because I can have restricted addresses with the same pub key and just different capabilities bits set. This means they will have the exact same signature. The uniqueness checks in the array rules do not account for this.

unlock_empty.go
```diff
@@ -0,0 +1,19 @@
...
+func (u *EmptyUnlock) Type() UnlockType {
+	return UnlockEmpty
+}
+
+func (u *EmptyUnlock) Size() int {
+	return serializer.SmallTypeDenotationByteSize
+}
+
+func (u *EmptyUnlock) WorkScore(_ *WorkScoreStructure) (WorkScore, error) {
+	return 0, nil
```
alexsporn, line 18: This is wrong, it costs at least 1 byte

unlock_test.go
```diff
@@ -32,6 +32,11 @@ func TestUnlock_DeSerialize(t *testing.T) {
 			source: tpkg.RandNFTUnlock(),
 			target: &iotago.NFTUnlock{},
 		},
+		{
+			name:   "ok - Mutli",
```
alexsporn: typo

alexsporn:
This new Unlock type brings funny new double-spends without actually double spending. I can take the same transaction and remove an optional address out of the multi-unlock so that it still reaches the threshhold and broadcast the tx again. Since the Tx has a different hash it will create different outputs.

unlock.go
muXxer: This was fixed by removing all existing restricted address types, except the multi address.
Now the uniqueness of addresses/pubkeys inside the multi unlock should be enforced again.

muXxer:
> This new Unlock type brings funny new double-spends without actually double spending. I can take the same transaction and remove an optional address out of the multi-unlock so that it still reaches the threshhold and broadcast the tx again. Since the Tx has a different hash it will create different outputs.

We should discuss this with the team, and handle it in another PR.

unlock.go
muXxer: I re-introduced the RestrictedEd25519 address again, so I had to solve this in a different way.
I added a `PubKeyHash` function to the address interface and added a custom validator that checks the uniqueness of these pubKeyHashes in the Address slice.

address_restricted_ed25519.go
```diff
@@ -30,8 +30,16 @@ func (addr *RestrictedEd25519Address) VBytes(rentStruct *RentStructure, _ VBytes
 	return rentStruct.VBFactorData.Multiply(VBytes(addr.Size()))
 }
 
+func (addr *RestrictedEd25519Address) ID() []byte {
+	return lo.PanicOnErr(CommonSerixAPI().Encode(context.TODO(), addr))
```
PhilippGackstatter, line 34: General nit: We should reserve "TODO" for when there's an actual TODO. Searching the code for TODOs will always show this as a hit. If it's not used here we should rename this to `None` or `Fake`, or something else that indicates that.

vm/vm.go
```diff
@@ -213,32 +213,83 @@ type UnlockedIdentities map[string]*UnlockedIdentity
...
+	if len(ident.Addresses) != len(multiUnlock.Unlocks) {
+		return ierrors.Wrapf(iotago.ErrMultiAddressAndUnlockLengthDoesNotMatch, "input %d has a multi address (%T) but the amount of addresses does not match the unlocks %d != %d", inputIndex, ident, len(ident.Addresses), len(multiUnlock.Unlocks))
+	}
+
+	var cumulativeUnlockedWeight uint16
+	for subIndex, unlock := range multiUnlock.Unlocks {
+		switch unlock.(type) {
+		case *iotago.EmptyUnlock:
+			// empty unlocks are simply skipped. they are used to maintain correct relationship between
+			// addresses and signatures if the signer doesn't know the particular signature.
```
PhilippGackstatter:
```suggestion
			// Empty unlocks are simply skipped. They are used to maintain correct index relationship between
			// addresses and signatures if the signer doesn't know the signature of another signer.
```

Maybe also add this info to the EmptyUnlock struct where it's more visible?

unlock.go
muXxer: Reverted the `PublicKeyHash` function changes. It's now a type switch in the validation.

address.go
```diff
@@ -83,6 +81,9 @@ type Address interface {
 	// Type returns the type of the address.
 	Type() AddressType
 
+	// ID returns the address ID.
+	ID() []byte
```
PhilippGackstatter:
```suggestion
    // ID returns the address ID, which is the concatenation of type prefix and raw address.
	ID() []byte
```

Something more helpful here would be great, because "ID" is super generic.

Do we still need a dedicated `Key` function in the interface? Can a caller not just use `string(addr.ID())` instead? 

Edit: I saw the Key used a bunch of times so perhaps it's convenient enough to keep it. Your call.

address_restricted_ed25519.go
```diff
@@ -30,8 +30,12 @@ func (addr *RestrictedEd25519Address) VBytes(rentStruct *RentStructure, _ VBytes
...
+}
+
 func (addr *RestrictedEd25519Address) Key() string {
-	return string(lo.PanicOnErr(CommonSerixAPI().Encode(context.TODO(), addr)))
+	return string(addr.ID())
 }
```
PhilippGackstatter, line 39: So now the key contains the capabilities, is that intended?

api_common.go
```diff
@@ -3,9 +3,76 @@ package iotago
...
+				addrWithoutTypeAndCapabilities = addr.PubKeyHash[:]
+			default:
+				return ierrors.Wrapf(ErrUnknownAddrType, "address with index %d has an unknown address type (%T) inside a multi address", idx, addr)
+			}
+
+			// we need to check for uniqueness of the address, but instead of the whole serialized address, we need to ignore
+			// different address types or capabilities, that might result in the same signature.
+			addrString := string(addrWithoutTypeAndCapabilities)
+			if j, has := addrSet[addrString]; has {
+				return ierrors.Wrapf(serializer.ErrArrayValidationViolatesUniqueness, "element %d and %d are duplicates", j, idx)
```
PhilippGackstatter:
```suggestion
				return ierrors.Wrapf(serializer.ErrArrayValidationViolatesUniqueness, "addresses with indices %d and %d in multi address are duplicates", j, idx)
```

Nit: More helpful/precise error message.

api_common.go
```diff
@@ -3,9 +3,76 @@ package iotago
...
+			}
+			addrSet[addrString] = idx
+
+			// check for minimum address weight
+			if address.Weight < 1 {
```
PhilippGackstatter:
```suggestion
			if address.Weight == 0 {
```

Nit: Seems weird to check for < 1 on a uint.

unlock.go
```diff
@@ -115,52 +131,147 @@ type ReferentialUnlock interface {
...
 				break
 			}
-			// must reference a sig unlock here
-			if _, has := seenSigUnlocks[x.Ref()]; !has {
-				return ierrors.Wrapf(ErrReferentialUnlockInvalid, "%d references non existent unlock %d", index, x.Ref())
+
+			// must reference a sig or multi unlock here
+			_, hasSigUnlock := seenSigUnlocks[unlock.Ref()]
+			_, hasMultiUnlock := seenMultiUnlocks[unlock.Ref()]
+			if !hasSigUnlock && !hasMultiUnlock {
```
PhilippGackstatter, line 203:
```suggestion
			_, isSigUnlock := seenSigUnlocks[unlock.Ref()]
			_, isMultiUnlock := seenMultiUnlocks[unlock.Ref()]
			if !isSigUnlock && !isMultiUnlock {
```

Nit: I feel like "is" would be more readable than "has" here.

unlock.go
```diff
@@ -115,52 +131,147 @@ type ReferentialUnlock interface {
...
-//  2. ReferenceUnlock(s) reference a previous SignatureUnlock
+//  1. SignatureUnlock(s) are unique (compared by public key)
+//     - SignatureUnlock(s) inside different MultiUnlock(s) don't need to be unique,
+//     as long as there is no equal SignatureUnlock(s) outside of a MultiUnlock(s).
+//  2. ReferenceUnlock(s) reference a previous SignatureUnlock or MultiUnlock
 //  3. Following through AccountUnlock(s), NFTUnlock(s) refs results to a SignatureUnlock
+//  4. EmptyUnlock(s) are only used inside of MultiUnlock(s)
+//  5. MultiUnlock(s) are not nested
+//  6. MultiUnlock(s) are unique
+//  7. ReferenceUnlock(s) to MultiUnlock(s) are not nested in MultiUnlock(s)
```
PhilippGackstatter, line 156: Nit: It's easier to read if you always use plural :).

unlock.go
```diff
@@ -115,52 +131,147 @@ type ReferentialUnlock interface {
...
+
+					// add the pubkey to "seenSigBlockPubkeyBytesInMultiUnlocks", so we can check that pubkeys from a multi unlock are not reused in a normal SignatureUnlock
+					seenSigBlockPubkeyBytesInMultiUnlocks[string(sigBlockPubKeyBytes)] = index
+
+				case ReferentialUnlock:
+					if prevRef := seenRefUnlocks[subUnlock.Ref()]; prevRef != nil {
+						if !subUnlock.Chainable() {
+							return ierrors.Wrapf(ErrReferentialUnlockInvalid, "%d.%d references existing referential unlock %d but it does not support chaining", index, subIndex, subUnlock.Ref())
+						}
+						// we don't set the index here in "seenRefUnlocks" because there is no concept of reference unlocks inside of multi unlocks
```
PhilippGackstatter:
```suggestion
						// we don't set the index here in "seenRefUnlocks" because it's not possible to reference an unlock within a multi unlock
```

Would this be clearer perhaps?

unlock.go
```diff
@@ -115,52 +131,147 @@ type ReferentialUnlock interface {
...
+
+						continue
+					}
+					// must reference a sig unlock here
+					// we don't check for "seenMultiUnlocks" here because we don't want to nest "reference unlocks to multi unlocks" in multi unlocks
+					if _, has := seenSigUnlocks[subUnlock.Ref()]; !has {
+						return ierrors.Wrapf(ErrReferentialUnlockInvalid, "%d.%d references non existent unlock %d", index, subIndex, subUnlock.Ref())
+					}
+					// we don't set the index here in "seenRefUnlocks" because there is no concept of reference unlocks inside of multi unlocks
```
PhilippGackstatter: This line appears twice, remove one of them.

address_ed25519.go
```diff
@@ -34,8 +34,12 @@ func (addr *Ed25519Address) VBytes(rentStruct *RentStructure, _ VBytesFunc) VByt
 	return rentStruct.VBFactorData.Multiply(VBytes(addr.Size()))
 }
 
+func (addr *Ed25519Address) ID() []byte {
```
alexsporn, line 37: maybe also use the ID() for the String() method

address_multi.go
```diff
@@ -0,0 +1,195 @@
...
+type MultiAddress struct {
+	Addresses           AddressesWithWeight        `serix:"0,mapKey=addresses"`
+	Threshold           uint16                     `serix:"1,mapKey=threshold"`
+	AllowedCapabilities AddressCapabilitiesBitMask `serix:"2,mapKey=allowedCapabilities,lengthPrefixType=uint8,maxLen=1"`
+}
+
+func (addr *MultiAddress) Clone() Address {
+	cpy := &MultiAddress{
+		Addresses: make(AddressesWithWeight, 0, len(addr.Addresses)),
+		Threshold: addr.Threshold,
```
alexsporn: this is not copying the `AllowedCapabilities`, so either remove that all from this PR in favor of the other one or fix it in this

api_common.go
```diff
@@ -3,9 +3,76 @@ package iotago
...
 
+	"github.com/iotaledger/hive.go/ierrors"
+	"github.com/iotaledger/hive.go/serializer/v2"
 	"github.com/iotaledger/hive.go/serializer/v2/serix"
 )
 
+var (
+	// the addresses need to be unique and lexically ordered to calculate a deterministic bech32 address for a MultiAddress.
+	// HINT: the uniqueness is checked within a custom validator function, which is on MultiAddress level.
+	addressesWithWeightV3ArrRules = &serix.ArrayRules{
```
alexsporn: remove `V3` from the name, since this is common api

alexsporn (approved):

vm/vm.go
```diff
@@ -487,18 +526,63 @@ func unlockOutput(vmParams *Params, output iotago.Output, inputIndex uint16) err
...
 
+		default:
+			return ierrors.Wrapf(iotago.ErrInvalidInputUnlock, "input %d has a direct unlockable address (%T) but its corresponding unlock is of type %T", inputIndex, owner, unlock)
+		}
+
+	case *iotago.MultiAddress:
+		switch uBlock := unlock.(type) {
+		case iotago.ReferentialUnlock:
+			if uBlock.Chainable() || !uBlock.SourceAllowed(ownerIdent) {
+				return ierrors.Wrapf(iotago.ErrInvalidInputUnlock, "input %d has none chain address of %s but its corresponding unlock is of type %s", inputIndex, owner.Type(), unlock.Type())
```
PhilippGackstatter:
```suggestion
				return ierrors.Wrapf(iotago.ErrInvalidInputUnlock, "input %d is a chain address of type %s but its corresponding unlock is of type %s", inputIndex, owner.Type(), unlock.Type())
```

I know this error message was copied but it's hard to understand. Not sure if this is much better, but at least a bit.

If I understand correctly, this is the case when I use a chain address and match it with an incorrect unlock, e.g. I try to unlock an NFT Address with an Account Unlock or an Account Address with an NFT Unlock.

If you accept this suggestion, also update the other occurence.

vm/stardust/vm_stardust_test.go
```diff
@@ -1264,6 +1264,1512 @@ func TestStardustTransactionExecution(t *testing.T) {
...
+				},
+				wantEncodeErr:  iotago.ErrMultiAddressThresholdInvalid,
+				wantExecuteErr: nil,
+			}
+		}(),
+
+		// fail - address weight == 0 (needs to be checked on output side)
+		func() test {
+			return test{
+				name:           "fail - address weight == 0 (needs to be checked on output side)",
```
PhilippGackstatter, line 2349: Nit: This case should have its own error type imo, because it's not really related to the threshold. I.e. don't use `ErrMultiAddressThresholdInvalid` but use a new type `ErrMultiAddressWeightInvalid` or something like that. That would be nicer for debugging, too.

vm/stardust/vm_stardust_test.go
```diff
@@ -1264,6 +1264,1512 @@ func TestStardustTransactionExecution(t *testing.T) {
...
+						},
+					}
+				},
+				wantEncodeErr:  nil,
+				wantExecuteErr: nil,
+			}
+		}(),
+
+		//
+		// Syntactical checks tests
```
PhilippGackstatter, line 2244: There are a bunch of syntactical tests in here that, if I'm not mistaken, could be moved to a simpler serialize/deserialize test, because some of the test stuff for other tests is not required for the syntactical checks. That would already clean up this test function a bit and make the whole test easier to understand. The whole thing is not exactly trivial, so keeping it somewhat minimal would be wonderful for future maintenance.

address_restricted_ed25519.go
muXxer: This is a golang internal function, and it's description is:
```
// TODO returns a non-nil, empty [Context]. Code should use context.TODO when
// it's unclear which Context to use or it is not yet available (because the
// surrounding function has not yet been extended to accept a Context
// parameter).
```

I also discussed with alex about that, because he introduced that. I personally would maybe go for "Background" instead.

address_restricted_ed25519.go
muXxer: That's intended.

unlock.go
muXxer: That's intended.

unlock.go
muXxer: Mhh I will keep it like this because there are several places in the code where we do this (complain to luca :lol:)

unlock.go
muXxer: isSigUnlock sounds more like a typecast. I would keep it as it is.

vm/stardust/vm_stardust_test.go
muXxer: I will fix this in another PR :thumbsup:

PhilippGackstatter (approved):

PR #503, opened 2023-09-14
Add Address Restriction & Implicit Account Creation Address checks

PhilippGackstatter:
# Description of change

Add checks for:
- Address restrictions in transactions.
- Implicit account creation addresses may not be present in any output type other than Basic Output or unlock condition type other than Address Unlock Condition.

Return Addresses in Expiration UC and Storage Deposit Return UC are not checked, because those do not directly affect the current transaction. As far as I can tell there is no way for someone to use the SDRUC to make a return transaction fail: A Basic Output with base tokens can always be received on any address.

## Type of change

Added a VM test to test restrictions.


api_v3.go
```diff
@@ -19,6 +19,14 @@ func must(err error) {
 	}
 }
 
+func implicitAccountCreationAddressValidator(address Address) error {
```
alexsporn: The naming of this validator is confusing, because you would think it validates it is an implicit account creation address, but it actually checks the opposite

address.go
```diff
@@ -12,6 +12,31 @@ import (
 var (
 	// ErrUnknownAddrType gets returned for unknown address types.
 	ErrUnknownAddrType = ierrors.New("unknown address type")
+	// ErrImplicitAccountCreationAddressInInvalidUnlockCondition gets returned when a Implicit Account Creation Address
+	// is placed in an unlock condition where it is disallowed.
+	ErrImplicitAccountCreationAddressInInvalidUnlockCondition = ierrors.New("implicit account creation address in unlock condition where it is disallowed")
+	// ErrImplicitAccountCreationAddressInInvalidOutput gets returned when a Implicit Account Creation Address
```
muXxer: Maybe use the exact name of the type, so we can better search for it in the code? :D

api_v3.go
PhilippGackstatter: Agreed, fixed it.

address.go
muXxer: I mean the comment `Implicit Account Creation Address`

vm/vm.go
```diff
@@ -725,3 +725,65 @@ func ExecFuncBalancedNativeTokens() ExecFunc {
...
+
+	if address.CannotReceiveDelegationOutputs() && output.Type() == iotago.OutputDelegation {
+		return iotago.ErrAddressCannotReceiveDelegationOutput
+	}
+
+	return nil
+}
+
+// Returns a func that validates that an address' restrictions are adhered to.
+// Does not validate the Return Address in Storage Deposit Return and Expiration UC.
```
muXxer: And where do we actually check that? Can't we add that check here as well?
IMO the transaction should better be invalid instead of locking funds in nimbus. WDYT?

muXxer: Ah I see, this is done in the additional validators?
Maybe we should add a comment here that this is done in the specific validators?

block_issuer_key_address_ed25519.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Type returns the BlockIssuerKeyType.
+func (key BlockIssuerKeyEd25519Address) Type() BlockIssuerKeyType {
+	return Ed25519BlockIssuerKeyAddress
+}
+
+// Size returns the size of the block issuer key when serialized.
+func (key BlockIssuerKeyEd25519Address) Size() int {
+	return serializer.SmallTypeDenotationByteSize + Ed25519AddressBytesLength
```
muXxer: Maybe call the Size() of the underlying address instead?

block_issuer_key_address_ed25519.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func (key BlockIssuerKeyEd25519Address) VBytes(_ *RentStructure, _ VBytesFunc) VBytes {
+	// type prefix + public key size
+	return serializer.SmallTypeDenotationByteSize + Ed25519AddressBytesLength
```
muXxer: Maybe base that on the result of the key.Size() function.

muXxer (approved):

PR #504, opened 2023-09-18
Indexer improvements for v3 nodeclient

alexsporn:
Added support for the indexer’s new outputs route and new unlockableByAddress query parameter to the nodeclient

### DIFF
--- a/nodeclient/indexer_models.go
+++ b/nodeclient/indexer_models.go
@@ -74,6 +74,12 @@ type IndexerNativeTokenParas struct {
 	MaxNativeTokenCount *uint32 `qs:"maxNativeTokenCount,omitempty"`
 }
 
+// IndexerUnlockableByAddressParas define address unlock related query parameters.
+type IndexerUnlockableByAddressParas struct {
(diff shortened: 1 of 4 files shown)


nodeclient/indexer_models.go
```diff
@@ -74,6 +74,12 @@ type IndexerNativeTokenParas struct {
 	MaxNativeTokenCount *uint32 `qs:"maxNativeTokenCount,omitempty"`
 }
 
+// IndexerUnlockableByAddressParas define address unlock related query parameters.
+type IndexerUnlockableByAddressParas struct {
+	// Bech32-encoded address that should be searched for.
+	UnlockableByAddressBech32 string `qs:"unlockableByAddress,omitempty"`
```
muXxer, line 80: why is it omitempty if that is the only criteria?

alexsporn: copy-paste

alexsporn: although, it needs to be omitempty, because there is always a default value in go

muXxer (approved):

PR #505, opened 2023-09-18
Unify restricted addresses in RestrictedAddress type

### LINKED ISSUE #347: Restricted addresses refactor

muXxer:
This PR unifies the different `RestrictedAddress` types into a single one which is basically a container for other address types.

A `RestrictedAddress` has a capabilities mask which can enable/disable the following features for that address:
- can receive `NativeTokens`
- can receive `Mana`
- can receive outputs with `TimelockUnlockCondition`
- can receive outputs with `ExpirationUnlockCondition`
- can receive outputs with `StorageDepositReturnUnlockCondition`
- can receive `AccountOutputs`
- can receive `NFTOutputs`
- can receive `DelegationOutputs`


address.go
```diff
@@ -196,7 +181,25 @@ func ParseBech32(s string) (NetworkPrefix, Address, error) {
 		return "", nil, serializer.ErrDeserializationNotEnoughData
 	}
 
-	addr, err := newAddress(addrData[0])
+	addrType := AddressType(addrData[0])
+
+	// check for invalid MultiAddresses in bech32 string
+	// MutliAddresses are hashed and can't be reconstructed via bech32
```
alexsporn: `Mutli` strikes again

address_restricted.go
```diff
@@ -0,0 +1,126 @@
...
+func (addr *RestrictedAddress) VBytes(rentStruct *RentStructure, _ VBytesFunc) VBytes {
+	return rentStruct.VBFactorData.Multiply(VBytes(addr.Size()))
+}
+
+func (addr *RestrictedAddress) ID() []byte {
+	addressID := addr.Address.ID()
+	capabilities := lo.PanicOnErr(CommonSerixAPI().Encode(context.TODO(), addr.AllowedCapabilities))
+
+	// prefix the ID of the underlying address with the AddressType, and append the capabilties
+	return append(append([]byte{byte(AddressRestricted)}, addressID...), capabilities...)
```
alexsporn: you are encoding by hand here, why not just encode it with serix?

muXxer: Because the ID() of a MultiAddress is not the serialized form, but the hash of it.

alexsporn: why not use `byteutils.ConcatBytes`

muXxer: Oh, didn't know that one. Fixed

alexsporn (approved):

PhilippGackstatter (approved):

PR #513, opened 2023-09-21
Add utilities in the transaction builder to allot the mana required to issue the block

muXxer:
This PR adds some utilities to calculate the available mana on input side, and to calculate the minimum mana required to issue a default block (4 strong parents) containing the transaction.

### DIFF
--- a/builder/output_builder.go
+++ b/builder/output_builder.go
@@ -527,15 +527,15 @@ func (trans *stakingTransition) FixedCost(fixedCost iotago.Mana) *stakingTransit
 }
 
 // StartEpoch sets the StartEpoch of iotago.StakingFeature.
-func (trans *stakingTransition) StartEpoch(slot iotago.EpochIndex) *stakingTransition {
-	trans.feature.StartEpoch = slot
+func (trans *stakingTransition) StartEpoch(epoch iotago.EpochIndex) *stakingTransition {
+	trans.feature.StartEpoch = epoch
 
 	return trans
 }
 
 // EndEpoch sets the EndEpoch of iotago.StakingFeature.
-func (trans *stakingTransition) EndEpoch(slot iotago.EpochIndex) *stakingTransition {
-	trans.feature.EndEpoch = slot
+func (trans *stakingTransition) EndEpoch(epoch iotago.EpochIndex) *stakingTransition {
+	trans.feature.EndEpoch = epoch
 
 	return trans
 }
--- a/builder/output_builder.go
+++ b/builder/output_builder.go
@@ -9,6 +9,7 @@ import (
 func NewBasicOutputBuilder(targetAddr iotago.Address, amount iotago.BaseToken) *BasicOutputBuilder {
 	return &BasicOutputBuilder{output: &iotago.BasicOutput{
 		Amount:       amount,
+		Mana:         0,
 		NativeTokens: iotago.NativeTokens{},
 		Conditions: iotago.BasicOutputUnlockConditions{
 			&iotago.AddressUnlockCondition{Address: targetAddr},
@@ -31,6 +32,13 @@ type BasicOutputBuilder struct {
 // Amount sets the base token amount of the output.
 func (builder *BasicOutputBuilder) Amount(amount iotago.BaseToken) *BasicOutputBuilder {
 	builder.output.Amount = amount
+
+	return builder
+}
+
+// Mana sets the mana of the output.
+func (builder *BasicOutputBuilder) Mana(mana iotago.Mana) *BasicOutputBuilder {
+	builder.output.Mana = mana
 
 	return builder
 }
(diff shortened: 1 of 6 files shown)


builder/output_builder.go
```diff
@@ -509,15 +527,15 @@ func (trans *stakingTransition) FixedCost(fixedCost iotago.Mana) *stakingTransit
...
+	trans.feature.StartEpoch = slot
 
 	return trans
 }
 
 // EndEpoch sets the EndEpoch of iotago.StakingFeature.
-func (trans *stakingTransition) EndEpoch(index iotago.EpochIndex) *stakingTransition {
-	trans.feature.EndEpoch = index
+func (trans *stakingTransition) EndEpoch(slot iotago.EpochIndex) *stakingTransition {
+	trans.feature.EndEpoch = slot
```
cyberphysic4l: should be epoch, not slot

builder/transaction_builder.go
```diff
@@ -81,9 +82,21 @@ func (b *TransactionBuilder) AddContextInput(contextInput iotago.Input) *Transac
 	return b
 }
 
-// AddAllotment adds the given allotment to the builder.
-func (b *TransactionBuilder) AddAllotment(allotment *iotago.Allotment) *TransactionBuilder {
-	b.transaction.Allotments = append(b.transaction.Allotments, allotment)
+// IncreaseAllotment adds or increases the given allotment to the builder.
+func (b *TransactionBuilder) IncreaseAllotment(accountID iotago.AccountID, value iotago.Mana) *TransactionBuilder {
+	// check if the allotment already exists and add the value on top
```
cyberphysic4l, line 91:
```suggestion
	if value == 0 {
        return b
    }
```

builder/transaction_builder.go
```diff
@@ -111,6 +124,137 @@ func (b *TransactionBuilder) AddTaggedDataPayload(payload *iotago.TaggedData) *T
...
+		if unboundManaInputs < storedMana {
+			return ierrors.New("not enough unbound mana available on the input side")
+		}
+		unboundManaInputs -= storedMana
+
+		return nil
+	}
+
+	// update the account bound mana balances if they exist and/or the onbound mana balance
+	updateUnboundAndAccountBoundManaBalances := func(accountID iotago.AccountID, storedMana iotago.Mana) error {
```
cyberphysic4l:
```suggestion
	updateUnboundAndAccountBoundManaBalances := func(accountID iotago.AccountID, accountBoundOutputMana iotago.Mana) error {
```
Rename this variable to make it clear we are subtracting account bound stuff from the output side

builder/transaction_builder.go
```diff
@@ -111,6 +124,137 @@ func (b *TransactionBuilder) AddTaggedDataPayload(payload *iotago.TaggedData) *T
...
+		return setBuildError(ierrors.Wrap(err, "failed to calculate the minimum required mana to issue the block"))
+	}
+
+	// subtract the minimum required mana to issue the block
+	if err := updateUnboundAndAccountBoundManaBalances(blockIssuerAccountID, minRequiredMana); err != nil {
+		return setBuildError(ierrors.Wrap(err, "failed to subtract the minimum required mana to issue the block"))
+	}
+
+	// allot the mana to the block issuer account (we increase the value, so we don't interfere with the already alloted value)
+	b.IncreaseAllotment(blockIssuerAccountID, minRequiredMana)
```
cyberphysic4l, line 245: Should we check if there is already allotment to this accountID and increase only up to the minRequiredMana rather than adding minRequiredMana extra?

builder/transaction_builder.go
```diff
@@ -128,6 +272,97 @@ func (b *TransactionBuilder) BuildAndSwapToBlockBuilder(signer iotago.AddressSig
...
+			excessBaseTokens, err := safemath.SafeSub(input.BaseTokenAmount(), minDeposit)
+			if err != nil {
+				return 0, 0, nil, ierrors.Wrap(err, "failed to calculate excessBaseTokens of the input")
+			}
+
+			potentialMana, err = protoParams.ManaDecayProvider().ManaGenerationWithDecay(excessBaseTokens, inputID.CreationSlot(), targetSlot)
+			if err != nil {
+				return 0, 0, nil, ierrors.Wrap(err, "failed to calculate potential mana generation and decay")
+			}
+		}
```
cyberphysic4l, line 300:
```suggestion
		if excessBaseTokens, _ := safemath.SafeSub(input.BaseTokenAmount(), minDeposit); excessBaseTokens > 0 {
			potentialMana, err = protoParams.ManaDecayProvider().ManaGenerationWithDecay(excessBaseTokens, inputID.CreationSlot(), targetSlot)
			if err != nil {
				return 0, 0, nil, ierrors.Wrap(err, "failed to calculate potential mana generation and decay")
			}
		}
```

cyberphysic4l (commented):
just a few changes requested

builder/transaction_builder.go
cyberphysic4l: This suggestion came out weird, but I obviously don't meant to delete the comment, just add the check for zero value

builder/transaction_builder.go
muXxer: I thinks its more clear for the reader if we do that check.

builder/transaction_builder.go
muXxer: I think its better to increase the allotment, as written in the comment.

cyberphysic4l (approved):

PR #516, opened 2023-09-22
Fix BlockIssuerKeys serialization, VBytes, Size and rename to `Ed25519BlockIssuerKeyPublicKey`

muXxer:


block_issuer_key_ed25519_address.go
```diff
@@ -1,32 +1,37 @@
...
-func (key BlockIssuerKeyEd25519Address) Size() int {
-	return key.Address.Size()
+func (key Ed25519AddressBlockIssuerKey) Size() int {
+	return serializer.SmallTypeDenotationByteSize + key.Address.Size()
 }
 
-func (key BlockIssuerKeyEd25519Address) VBytes(rentStructure *RentStructure, vbyteFunc VBytesFunc) VBytes {
-	return key.Address.VBytes(rentStructure, vbyteFunc)
+func (key Ed25519AddressBlockIssuerKey) VBytes(rentStructure *RentStructure, vbyteFunc VBytesFunc) VBytes {
+	return rentStructure.VBFactorData.Multiply(VBytes(serializer.SmallTypeDenotationByteSize)) + key.Address.VBytes(rentStructure, vbyteFunc)
```
PhilippGackstatter: Block Issuer Keys have their own rent structure weight, so we shouldn't use the data factor. That is currently being applied in the `BlockIssuerKeys` type, but it would be clearer if it was here, actually.

I think this should be:

```go
return rentStructure.VBFactorBlockIssuer.Multiply(VBytes(key.Size()))
```

We shouldn't delegate to `VBytes` of the address, because the block issuer key and the address type have different storage deposit rules, hence why `key.Size()` is better I think.

block_issuer_key_ed25519_pubkey.go
```diff
@@ -0,0 +1,43 @@
...
+	return BlockIssuerKeyEd25519PublicKey
+}
+
+// Size returns the size of the block issuer key when serialized.
+func (key Ed25519PublicKeyBlockIssuerKey) Size() int {
+	return serializer.SmallTypeDenotationByteSize + ed25519.PublicKeySize
+}
+
+func (key Ed25519PublicKeyBlockIssuerKey) VBytes(rentStruct *RentStructure, _ VBytesFunc) VBytes {
+	return rentStruct.VBFactorData.Multiply(VBytes(key.Size()))
```
PhilippGackstatter: Same here as for the other vbytes.

PR #517, opened 2023-09-22
Add creation slot to the TransactionID and therefore all OutputIDs

muXxer:
It would be much easier for mana calculations on client side to encode the creation slot of an output into the `OutputID`.
This way a client could calculate the generated and decay mana based on the ID.

Since the `OutputID` are based on the `TransactionID`, we can embed this information there.
It is based on the `CreationSlot` that is part of the `TransactionEssence`.

It also changes the `SlotIndex` and `EpochIndex` types to use 32bit instead of 64bit.


merklehasher/merkle_hasher_test.go
```diff
@@ -22,19 +22,20 @@ func TestMerkleHasher(t *testing.T) {
...
+	includedBlocks = append(includedBlocks, iotago.MustSlotIdentifierFromHexString("0x0bf5059875921e668a5bdf2c7fc4844592d2572bcd0668d2d6c52f5054e2d08300000000"))
+	includedBlocks = append(includedBlocks, iotago.MustSlotIdentifierFromHexString("0x6bf84c7174cb7476364cc3dbd968b0f7172ed85794bb358b0c3b525da1786f9f00000000"))
 
 	hasher := merklehasher.NewHasher[iotago.BlockID](crypto.BLAKE2b_256)
 	hash, err := hasher.HashValues(includedBlocks)
 	require.NoError(t, err)
 
-	expectedHash, err := hexutil.DecodeHex("0x1f943b38f71a984d154b7ffef0ff0481c644b3c796d3a451fbe7ebd909b1022c")
+	expectedHash, err := hexutil.DecodeHex("0x4a6ff2aca6a11554b6997cf91c31585d436235e7a45f6b4ea48648d6488f6726")
+	println(string(hexutil.EncodeHex(hash)))
```
karimodm: Remove

transaction.go
```diff
@@ -9,7 +9,7 @@ import (
 
 const (
 	// TransactionIDLength defines the length of a Transaction ID.
-	TransactionIDLength = IdentifierLength
+	TransactionIDLength = SlotIndexLength + IdentifierLength
```
karimodm: Invert to reflect the order as the slot is a suffix?

karimodm (approved):

PR #535, opened 2023-09-28
Remove OutputWithCreationSlot from VM

muXxer:
This PR removes the `OutputWithCreationSlot` from the VM. The necessary information about the creation slot is now part of the outputID anyway.


vm/types.go
```diff
@@ -6,41 +6,36 @@ import (
...
 	}
 
 	return outputs
 }
 
-type ChainOutputWithCreationSlot struct {
-	ChainID      iotago.ChainID
-	Output       iotago.ChainOutput
-	CreationSlot iotago.SlotIndex
+type ChainOutput struct {
```
cyberphysic4l: I think this should be renamed to ChainOutputWithOutputID. It is a little confusing when we use this type (vm.ChainOutput) for the inputs and iotago.ChainOutput on the output side in some places in the VM.

vm/stardust/vm_stardust.go
```diff
@@ -337,7 +337,7 @@ func accountBlockIssuerSTVF(input *vm.ChainOutputWithCreationSlot, next *iotago.
 	}
 
 	// AccountInStored
-	manaStoredAccount, err := manaDecayProvider.ManaWithDecay(current.Mana, input.CreationSlot, vmParams.WorkingSet.Tx.Essence.CreationSlot)
+	manaStoredAccount, err := manaDecayProvider.ManaWithDecay(current.Mana, input.OutputID.CreationSlotIndex(), vmParams.WorkingSet.Tx.Essence.CreationSlot)
```
cyberphysic4l: Shouldn't this be CreationSlot() rather than CreationSlotIndex()?

cyberphysic4l (approved):

PR #537, opened 2023-09-28
VBytes -> StorageScore

cyberphysic4l:
This PR renames VBytes to StorageScore.

We also change to using offsets instead of factors objects that incur extra recurring costs in the node. The offsets mean we apply extra storageScore for each such object rather than multiplying the object's bytes by a factor. This makes more sense as these offsets are independent of the object byte size.


rent.go
```diff
@@ -125,57 +125,59 @@ func NewRentStructure(rentParameters *RentParameters) *RentStructure {
...
-	vBFactorDummyBasicOutput := dummyBasicOutput.VBytes(rentStructure, nil)
-	vBFactorDummyAddress := dummyAddress.VBytes(rentStructure, nil)
-	rentStructure.VBFactorImplicitAccountCreationAddress = VByteCostFactor(vBFactorDummyAccountOutput - vBFactorDummyBasicOutput + vBFactorDummyAddress)
+	// set the storage score offset for implicit account creation addresses as
+	// the difference between the storage score of the dummy account and the storage
+	// score of the dummy basic output minus the storage score of the dummy address.
+	storageScoreAccountOutput := dummyAccountOutput.StorageScore(rentStructure, nil)
+	storageScoreBasicOutput := dummyBasicOutput.StorageScore(rentStructure, nil)
+	storageScoreAddress := dummyAddress.StorageScore(rentStructure, nil)
+	rentStructure.StorageScoreOffsetImplicitAccountCreationAddress = storageScoreAccountOutput - storageScoreBasicOutput + storageScoreAddress
```
muXxer: Maybe check for underflows here with safemath and panic?

rent.go
```diff
@@ -125,57 +125,59 @@ func NewRentStructure(rentParameters *RentParameters) *RentStructure {
...
+		return 0, ierrors.Wrapf(ErrStorageDepositNotCovered, "needed %d but only got %d", minDeposit, amount)
 	}
 
 	return minDeposit, nil
 }
 
 // MinDeposit returns the minimum deposit to cover a given object.
 func (r *RentStructure) MinDeposit(object NonEphemeralObject) BaseToken {
-	return BaseToken(r.VByteCost()) * BaseToken(object.VBytes(r, nil))
+	return r.StorageCost() * BaseToken(object.StorageScore(r, nil))
```
muXxer: Why don't we use safemath here?

rent.go
```diff
@@ -125,57 +125,59 @@ func NewRentStructure(rentParameters *RentParameters) *RentStructure {
...
 }
 
 // MinStorageDepositForReturnOutput returns the minimum renting costs for an BasicOutput which returns
 // a StorageDepositReturnUnlockCondition amount back to the origin sender.
 func (r *RentStructure) MinStorageDepositForReturnOutput(sender Address) BaseToken {
-	return BaseToken(r.VByteCost()) * BaseToken((&BasicOutput{Conditions: UnlockConditions[basicOutputUnlockCondition]{&AddressUnlockCondition{Address: sender}}, Amount: 0}).VBytes(r, nil))
+	return r.StorageCost() * BaseToken((&BasicOutput{Conditions: UnlockConditions[basicOutputUnlockCondition]{&AddressUnlockCondition{Address: sender}}, Amount: 0}).StorageScore(r, nil))
```
muXxer: Why don't we use safemath here?

muXxer (changes requested):

vm/stardust/vm_stardust.go
```diff
@@ -465,13 +466,12 @@ func accountBlockIssuerSTVF(input *vm.ChainOutputWithIDs, currentBlockIssuerFeat
...
-	if current.BaseTokenAmount() <= minDeposit {
-		excessBaseTokensAccount = 0
-	} else {
-		excessBaseTokensAccount = current.BaseTokenAmount() - minDeposit
+	minDeposit, err := rentStructure.MinDeposit(current)
+	if err != nil {
+		return err
 	}
+	// no need to check underflow error as the result of safe sub will be zero in that case
+	excessBaseTokensAccount, _ := safemath.SafeSub(current.BaseTokenAmount(), minDeposit)
```
muXxer: Better check for the error here and replace the excessBaseTokensAccount with 0 if you get the underflow error.

output.go
```diff
@@ -776,7 +776,10 @@ func OutputsSyntacticalDepositAmount(protoParams ProtocolParameters, rentStructu
 
 		// check whether the amount in the return condition allows the receiver to fulfill the storage deposit for the return output
 		if storageDep := output.UnlockConditionSet().StorageDepositReturn(); storageDep != nil {
-			minStorageDepositForReturnOutput := rentStructure.MinStorageDepositForReturnOutput(storageDep.ReturnAddress)
+			minStorageDepositForReturnOutput, err := rentStructure.MinStorageDepositForReturnOutput(storageDep.ReturnAddress)
+			if err != nil {
+				return ierrors.Wrapf(err, "output index %d", index)
```
muXxer: maybe extend the error message a bit, because we will only know about an overflow/underflow, and the output index. but not that it was about storage deposit calculation.

rent.go
```diff
@@ -140,7 +143,10 @@ func NewRentStructure(rentParameters *RentParameters) *RentStructure {
 // by examining the storage score of the object.
 // Returns the minimum deposit computed and an error if it is not covered by the base token amount of the object.
 func (r *RentStructure) CoversMinDeposit(object NonEphemeralObject, amount BaseToken) (BaseToken, error) {
-	minDeposit := r.MinDeposit(object)
+	minDeposit, err := r.MinDeposit(object)
+	if err != nil {
+		return 0, err
```
muXxer: Maybe extend the error message a bit here as well.

muXxer (changes requested):

muXxer (approved):

PR #545, opened 2023-10-01
BREAKING: Make NativeToken a feature and remove the slices (only one NT Type per Output)

muXxer:
We simplify the Stardust NativeTokens by doing the following changes:
- Only BasicOutput and FoundryOutput can hold NativeTokens
- NativeTokens are not a list/slice of NativeTokens anymore, but only one NativeToken Type per Output is allowed
- We remove the MaxNativeToken limits per Transaction
- FoundryOutput can only hold the NativeToken Type of the Foundry itself
- NativeToken is now a feature of the output, not an extra field


feat_native_token.go
```diff
@@ -0,0 +1,89 @@
...
+	// Uint256ByteSize defines the size of an uint256.
+	Uint256ByteSize = 32
+
+	// NativeTokenIDLength is the byte length of a NativeTokenID which is the same thing as a FoundryID.
+	NativeTokenIDLength = FoundryIDLength
+)
+
+var (
+	// ErrNativeTokenAmountLessThanEqualZero gets returned when a NativeToken.Amount is not bigger than 0.
+	ErrNativeTokenAmountLessThanEqualZero = ierrors.New("native token must be a value bigger than zero")
```
PhilippGackstatter:
```suggestion
	ErrNativeTokenAmountLessThanEqualZero = ierrors.New("native token amount must be greater than zero")
```

feat_native_token.go
```diff
@@ -0,0 +1,89 @@
...
+	// ErrNativeTokenSumExceedsUint256 gets returned when a NativeToken.Amount addition results in a value bigger than the max value of a uint256.
+	ErrNativeTokenSumExceedsUint256 = ierrors.New("native token sum exceeds max value of a uint256")
+	// ErrNativeTokenSumUnbalanced gets returned when two NativeTokenSum(s) are unbalanced.
+	ErrNativeTokenSumUnbalanced = ierrors.New("native token sums are unbalanced")
+	// ErrFoundryIDNativeTokenIDMismatch gets returned when a native token features exists in a foundry output but the IDs mismatch.
+	ErrFoundryIDNativeTokenIDMismatch = ierrors.New("native token feature ID in foundry output must match the foundry ID")
```
PhilippGackstatter:
```suggestion
	ErrFoundryIDNativeTokenIDMismatch = ierrors.New("native token ID in foundry output must match the foundry ID")
```

Otherwise this reads like the native token feature has an ID.

PhilippGackstatter (approved):

PR #547, opened 2023-10-03
Add mutation proof and make TransactionIDs serializable with serix

jonastheis:
This PR adds a method to create a `MutationProof` from `Roots` and makes `TransactionID` serializable with serix by adding a rule for it.


hmoog (approved):
LGTM

PR #551, opened 2023-10-03
It begins with stardust...

muXxer:
...and ends in a super-nova, enriching the cosmos with the building blocks of new stars and planets.

We need a new name for our UTXO VM. :)


PhilippGackstatter (approved):
:star2:

PR #557, opened 2023-10-05
Code generation and type safety

jonastheis:
This PR introduces code generation with templates for our basic types so that we don't have all of them being aliases. This prevents "pollution" of the basic type's namespace and guarantees type safety at compile time thus preventing coding mistakes down the line.


block_id.gen.go
```diff
@@ -0,0 +1,201 @@
...
+	BlockIDAliases = make(map[BlockID]string)
+
+	// BlockIDAliasesMutex is the mutex that is used to synchronize access to the previous map.
+	BlockIDAliasesMutex = sync.RWMutex{}
+)
+
+// RegisterAlias allows to register a human-readable alias for the Identifier which will be used as a replacement for
+// the String method.
+func (b BlockID) RegisterAlias(alias string) {
+	BlockIDAliasesMutex.Lock()
```
muXxer: Seems like the "firstLower .Name" doesn't work correctly?

muXxer (changes requested):
Looks good to me, except the lowercase for the mutexes.

block_id.gen.go
jonastheis: It works but I didn't call it for the `slot_identifier` template

muXxer (approved):

PR #559, opened 2023-10-05
Align naming: structure to parameters

jkrvivian:
* manaStructure -> manaParameters
* workScoreStructure -> workScoreParameters


api.go
```diff
@@ -112,8 +112,8 @@ type ProtocolParameters interface {
 	Bech32HRP() NetworkPrefix
 	// RentStructure defines the rent structure used by given node/network.
 	RentParameters() *RentParameters
-	// WorkScoreStructure defines the work score structure used by the given network.
-	WorkScoreStructure() *WorkScoreStructure
+	// WorkScoreParameters defines the work score structure used by the given network.
```
muXxer: please change the comment as well.

api_protocol_parameters.go
```diff
@@ -11,8 +11,8 @@ type basicProtocolParameters struct {
 
 	// RentStructure defines the rent structure used by given node/network.
 	RentParameters RentParameters `serix:"3,mapKey=rentParameters"`
-	// WorkScoreStructure defines the work score structure used by given node/network.
-	WorkScoreStructure WorkScoreStructure `serix:"4,mapKey=workScoreStructure"`
+	// WorkScoreParameters defines the work score structure used by given node/network.
```
muXxer: please change the comment as well.

muXxer (changes requested):

muXxer (approved):

alexsporn:
Some of the objects (Blocks and Transactions) have now an API inside and could use that one to calculate the WorkScore.
Also the TransactionBuilder does not require having it as a parameter because it contains a valid API inside.

We can do these changes in a separate PR if you don't want to do them here.

cyberphysic4l:
A further change we could make here is to do something similar to what we do with the StorageScoreParameter/StorageScoreStructure. The Parameters are what we set, and the Structure contains the Parameters plus some other values derived from the parameters, and this lives in the API. MaxBlockWorkScore, which is derived from the WorkScoreParameters would live inside this WorkScoreStructure instead of being on its own in the API in this case.

This way of doing things requires a little bit of bloating of the API but a nicer way to organise imo.

PR #565, opened 2023-10-11
WIP: Split account output into account and anchor output

muXxer:
WIP:
- the tests need to be adapted
- All tests that were account related should be copied for the anchor output as well.


muXxer:
Superseded by #589

PR #569, opened 2023-10-13
Add AllotAllMana to txBuilder

daria305:
This PR is adding function `AllotAllMana` to `transactionBuilder`, which is allotting all potential mana to given account and does not check for minimum allotment requirements based on RMC.

It is needed for PR: https://github.com/iotaledger/iota-core/pull/386

### DIFF
--- a/builder/transaction_builder.go
+++ b/builder/transaction_builder.go
@@ -146,42 +146,66 @@ func (b *TransactionBuilder) AllotRequiredManaAndStoreRemainingManaInOutput(targ
 		return b
 	}
 
-	minManalockedSlot := b.transaction.CreationSlot + 2*b.api.ProtocolParameters().MaxCommittableAge()
-
-	// check if the output is locked for a certain time to an account.
-	hasManalockCondition := func(output iotago.Output) (iotago.AccountID, bool) {
-		if !output.UnlockConditionSet().HasTimelockUntil(minManalockedSlot) {
-			return iotago.EmptyAccountID, false
-		}
-
-		unlockAddress := output.UnlockConditionSet().Address()
-		if unlockAddress == nil {
-			return iotago.EmptyAccountID, false
-		}
-
-		if unlockAddress.Address.Type() != iotago.AddressAccount {
-			return iotago.EmptyAccountID, false
-		}
-		//nolint:forcetypeassert // we can safely assume that this is an AccountAddress
-		accountAddress := unlockAddress.Address.(*iotago.AccountAddress)
-
-		return accountAddress.AccountID(), true
+	if storedManaOutputIndex >= len(b.transaction.Outputs) {
+		return setBuildError(ierrors.Errorf("given storedManaOutputIndex does not exist: %d", storedManaOutputIndex))
 	}
 
+	// calculate the minimum required mana to issue the block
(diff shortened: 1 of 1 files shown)


builder/transaction_builder.go
```diff
@@ -146,42 +146,62 @@ func (b *TransactionBuilder) AllotRequiredManaAndStoreRemainingManaInOutput(targ
...
+		output.Mana += unboundManaInputsLeftoverBalance
+	default:
+		return setBuildError(ierrors.Wrapf(iotago.ErrUnknownOutputType, "output type %T does not support stored mana", output))
+	}
+
+	return b
+}
 
-		return accountAddress.AccountID(), true
+// AllotAllMana allots all loose mana to the provided account, even if alloted value is less than required paymnet.
```
muXxer:
```suggestion
// AllotAllMana allots all available mana to the provided account, even if the alloted value is less than the minimum required mana value to issue the block.
```

builder/transaction_builder.go
```diff
@@ -146,42 +146,62 @@ func (b *TransactionBuilder) AllotRequiredManaAndStoreRemainingManaInOutput(targ
...
+	minRequiredMana, err := b.MinRequiredAllotedMana(b.api.ProtocolParameters().WorkScoreStructure(), rmc, blockIssuerAccountID)
+	if err != nil {
+		return setBuildError(ierrors.Wrap(err, "failed to calculate the minimum required mana to issue the block"))
+	}
 
-		unlockAddress := output.UnlockConditionSet().Address()
-		if unlockAddress == nil {
-			return iotago.EmptyAccountID, false
-		}
+	unboundManaInputsLeftoverBalance, err := b.calculateMaximumPossibleAllotment(targetSlot, minRequiredMana, blockIssuerAccountID)
```
muXxer: I find this function name confusing. Why is this about the maximum possible allotment, and not the leftover available mana?

builder/transaction_builder.go
```diff
@@ -205,70 +230,73 @@ func (b *TransactionBuilder) AllotRequiredManaAndStoreRemainingManaInOutput(targ
...
 			}
 
 		default:
 			// check if the output locked mana to a certain account
-			if accountID, isManaLocked := hasManalockCondition(output); isManaLocked {
-				if err := updateUnboundAndAccountBoundManaBalances(accountID, output.StoredMana()); err != nil {
-					return setBuildError(ierrors.Wrap(err, "failed to subtract the stored mana on the outputs side"))
+			if accountID, isManaLocked := b.hasManalockCondition(output); isManaLocked {
+				if err = updateUnboundAndAccountBoundManaBalances(accountID, output.StoredMana()); err != nil {
+					return 0, ierrors.Wrap(err, "failed to subtract the stored mana on the outputs side because of lock")
```
muXxer:
```suggestion
					return 0, ierrors.Wrap(err, "failed to subtract the stored mana on the outputs side because of an existing mana lock condition")
```

muXxer: Why is that even a problem with the lock? Do we need to mention that?

builder/transaction_builder.go
daria305: yup, that's better name

builder/transaction_builder.go
daria305: added just to differentiate from the next error, but its not actually because of, I changed it to "failed to subtract the stored mana on the outputs side, while checking locked mana"

muXxer (approved):

PR #571, opened 2023-10-17
Remove dummy address in Implicit Account Creation Address storage score calculation

PhilippGackstatter:
# Description of change

Remove dummy address in Implicit Account Creation Address storage score calculation. Due to the recent changes in the storage score, the storage score of a plain Ed25519 Address is 0, so it no longer affects the calculation like it did before. Previously we had to subtract the Ed25519 Address because it added an extra vbyte cost on top of what the basic output itself had. With the change that the output's serialized size is only taken into account at the top-level, this no longer matters. The size of an Ed25519 Address and an Implicit Account Creation Address is the same, so the storage score of the data part of the equation is the same for a regular Basic Output and an Implicit Account. Hence now the implicit account creation address storage score offset can simply be that of a minimal block issuer account minus that of a minimal basic output.

## Type of change

Existing tests pass.

### DIFF
--- a/rent.go
+++ b/rent.go
@@ -53,7 +53,7 @@ type RentStructure struct {
 	RentParameters *RentParameters
 	// The storage score that a minimal block issuer account needs to have minus the wrapping Basic Output.
 	// Since this value is used for implicit account creation addresses, this value plus the wrapping
-	// Basic Output (the Implicit Account Creation Address is contained in) results in the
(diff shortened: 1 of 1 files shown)


PhilippGackstatter:
Closed in favor of leaving the correct equation in place, even though the score of the dummy address is 0, for now anyway.

PR #579, opened 2023-10-23
Cleanup ProtocolParameters interface

muXxer:
We still have functions that always reinitialize `ManaDecayProvider` and `TimeProvider`. That should be part of the `API`, but not the `ProtocolParameters`.


hmoog (approved):
LGTM

PR #583, opened 2023-10-24
Do not use AvailableManaResult directly in calculateAvailableManaLeftover

muXxer:
A small PR that reverts some changes from the last PR.

### DIFF
--- a/builder/transaction_builder.go
+++ b/builder/transaction_builder.go
@@ -204,18 +204,21 @@ func (b *TransactionBuilder) calculateAvailableManaLeftover(targetSlot iotago.Sl
 		return 0, ierrors.Wrap(err, "failed to calculate the available mana on input side")
 	}
 
+	unboundManaInputs := availableManaInputs.UnboundMana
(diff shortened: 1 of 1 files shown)


alexsporn (dismissed):

builder/transaction_builder.go
```diff
@@ -204,18 +204,21 @@ func (b *TransactionBuilder) calculateAvailableManaLeftover(targetSlot iotago.Sl
 		return 0, ierrors.Wrap(err, "failed to calculate the available mana on input side")
 	}
 
+	unboundManaInputs := availableManaInputs.UnboundMana
+	accountBoundManaInputs := availableManaInputs.AccountBoundMana
```
alexsporn, line 208: This is not making a copy

muXxer:
This PR was meant to show that we don't actually meant to modify the `*AvailableManaResult`, but only it's result is used in further calculations. But actually it doesn't matter or not. I don't care. If that makes it even more unclear, I just close the PR.

PR #589, opened 2023-10-26
Add AnchorOutput

### LINKED ISSUE #405: Reduce AccountOutput in favor of AnchorOutput
To simplify handling of AccountOutputs, we want to move state governor transitions fields into a dedicated `AnchorOutput`, effectively splitting concerns between ISC and Account use-cases.

### LINKED ISSUE #410: Adapt INX-Indexer to newest nova VM
```[tasklist]
### Tasks
- [ ] Add support for AnchorOutputs
- [x] Adapt NativeTokens and make them queryable
- [ ] Check if we can do some easy balance calculations
- [x] Check data types (SlotIndex 32bit, payload type 8bit etc)
- [x] Add unit-tests to verify indexer is properly indexing stuff
```

muXxer:
This PR splits the current `AccountOutput` functionality into `AccountOutput` and `AnchorOutput`.

### DIFF
--- a/api_v3.go
+++ b/api_v3.go
@@ -32,8 +32,8 @@ func disallowImplicitAccountCreationAddress(address Address) error {
 
 var (
 	basicOutputV3UnlockCondArrRules = &serix.ArrayRules{
-		Min: 1,
-		Max: 4,
+		Min: 1, // Min: AddressUnlockCondition
+		Max: 4, // Max: AddressUnlockCondition, StorageDepositReturnUnlockCondition, TimelockUnlockCondition, ExpirationUnlockCondition
 		MustOccur: serializer.TypePrefixes{
 			uint32(UnlockConditionAddress): struct{}{},
 		},
@@ -42,15 +42,43 @@ var (
 			serializer.ArrayValidationModeAtMostOneOfEachTypeByte,
 	}
 	basicOutputV3FeatBlocksArrRules = &serix.ArrayRules{
-		Min: 0,
-		Max: 4,
+		Min: 0, // Min: -
+		Max: 4, // Max: SenderFeature, MetadataFeature, TagFeature, NativeTokenFeature
 		ValidationMode: serializer.ArrayValidationModeNoDuplicates |
 			serializer.ArrayValidationModeLexicalOrdering |
 			serializer.ArrayValidationModeAtMostOneOfEachTypeByte,
 	}
 
 	accountOutputV3UnlockCondArrRules = &serix.ArrayRules{
-		Min: 2, Max: 2,
+		Min: 1, // Min: AddressUnlockCondition
+		Max: 1, // Max: AddressUnlockCondition
+		MustOccur: serializer.TypePrefixes{
+			uint32(UnlockConditionAddress): struct{}{},
+		},
+		ValidationMode: serializer.ArrayValidationModeNoDuplicates |
+			serializer.ArrayValidationModeLexicalOrdering |
+			serializer.ArrayValidationModeAtMostOneOfEachTypeByte,
+	}
+
+	accountOutputV3FeatBlocksArrRules = &serix.ArrayRules{
+		Min: 0, // Min: -
+		Max: 4, // Max: SenderFeature, MetadataFeature, BlockIssuerFeature, StakingFeature
+		ValidationMode: serializer.ArrayValidationModeNoDuplicates |
+			serializer.ArrayValidationModeLexicalOrdering |
+			serializer.ArrayValidationModeAtMostOneOfEachTypeByte,
+	}
+
+	accountOutputV3ImmFeatBlocksArrRules = &serix.ArrayRules{
+		Min: 0, // Min: -
+		Max: 2, // Max: IssuerFeature, MetadataFeature
+		ValidationMode: serializer.ArrayValidationModeNoDuplicates |
+			serializer.ArrayValidationModeLexicalOrdering |
+			serializer.ArrayValidationModeAtMostOneOfEachTypeByte,
+	}
+
+	anchorOutputV3UnlockCondArrRules = &serix.ArrayRules{
+		Min: 2, // Min: StateControllerAddressUnlockCondition, GovernorAddressUnlockCondition
+		Max: 2, // Max: StateControllerAddressUnlockCondition, GovernorAddressUnlockCondition
 		MustOccur: serializer.TypePrefixes{
 			uint32(UnlockConditionStateControllerAddress): struct{}{},
 			uint32(UnlockConditionGovernorAddress):        struct{}{},
@@ -60,24 +88,25 @@ var (
 			serializer.ArrayValidationModeAtMostOneOfEachTypeByte,
 	}
 
-	accountOutputV3FeatBlocksArrRules = &serix.ArrayRules{
-		Min: 0,
-		Max: 4,
+	anchorOutputV3FeatBlocksArrRules = &serix.ArrayRules{
+		Min: 0, // Min: -
+		Max: 2, // Max: SenderFeature, MetadataFeature
 		ValidationMode: serializer.ArrayValidationModeNoDuplicates |
 			serializer.ArrayValidationModeLexicalOrdering |
 			serializer.ArrayValidationModeAtMostOneOfEachTypeByte,
 	}
 
-	accountOutputV3ImmFeatBlocksArrRules = &serix.ArrayRules{
-		Min: 0,
-		Max: 2,
+	anchorOutputV3ImmFeatBlocksArrRules = &serix.ArrayRules{
+		Min: 0, // Min: -
+		Max: 2, // Max: IssuerFeature, MetadataFeature
 		ValidationMode: serializer.ArrayValidationModeNoDuplicates |
 			serializer.ArrayValidationModeLexicalOrdering |
 			serializer.ArrayValidationModeAtMostOneOfEachTypeByte,
 	}
 
 	foundryOutputV3UnlockCondArrRules = &serix.ArrayRules{
-		Min: 1, Max: 1,
+		Min: 1, // Min: ImmutableAccountUnlockCondition
+		Max: 1, // Max: ImmutableAccountUnlockCondition
 		MustOccur: serializer.TypePrefixes{
 			uint32(UnlockConditionImmutableAccount): struct{}{},
 		},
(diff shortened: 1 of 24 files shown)


vm/nova/vm.go
```diff
@@ -301,97 +307,15 @@ func accountStateChangeValid(vmParams *vm.Params, input *vm.ChainOutputWithIDs,
...
-
-	return accountStateSTVF(vmParams, input, next)
-}
-
-func accountGovernanceSTVF(vmParams *vm.Params, input *vm.ChainOutputWithIDs, next *iotago.AccountOutput) error {
-	//nolint:forcetypeassert // we can safely assume that this is an AccountOutput
-	current := input.Output.(*iotago.AccountOutput)
-
-	switch {
-	case current.Amount != next.Amount:
```
muXxer, line 316: @cyberphysic4l @PhilippGackstatter please have a closer look at those change, because I had to merge the governor and state transition logic here.

vm/nova/vm.go
```diff
@@ -447,14 +372,23 @@ func accountBlockIssuerSTVF(vmParams *vm.Params, input *vm.ChainOutputWithIDs, c
 	manaIn := vmParams.WorkingSet.TotalManaIn
 	manaOut := vmParams.WorkingSet.TotalManaOut
 
+	// add the claimed mana rewards of the account in question
+	if manaReward, isClaiming := vmParams.WorkingSet.Rewards[next.ChainID()]; isClaiming {
```
muXxer: @cyberphysic4l @PhilippGackstatter please have a closer look here. I think that was missing before, because we never checked the staking claiming together with the block issuer feature correctly.

muXxer: Is this maybe missing somewhere else as well in the codebase?

api_v3.go
```diff
@@ -450,6 +483,34 @@ func V3API(protoParams ProtocolParameters) API {
 		must(api.RegisterInterfaceObjects((*accountOutputImmFeature)(nil), (*MetadataFeature)(nil)))
 	}
 
+	{
+		must(api.RegisterTypeSettings(AnchorOutput{}, serix.TypeSettings{}.WithObjectType(uint8(OutputAnchor))))
+		must(api.RegisterValidators(AnchorOutput{}, nil, func(ctx context.Context, anchor AnchorOutput) error {
+			return anchor.syntacticallyValidate()
```
PhilippGackstatter, line 489: Not directly related to your PR, but I introduced this pattern for implicit account address checks while not realizing that we already have `Transaction.syntacticallyValidate` which calls syntactic validation functions on all the contained outputs. I think we should reconcile the two (not in this PR), so the syntactic checks are not dispersed across different functions, but I'm not sure which pattern is better. Having everything related to syntactic checks in the api file here (where also the max and min len of arrays is defined, and things like that) makes it nicely contained and easy to find.

I did a quick test and it seems like we can register a validator for interfaces as well, so that we can, for instance, also register a validator for the storage deposit which would operate on any output.

```go
		must(api.RegisterValidators((*TxEssenceOutput)(nil), nil, func(ctx context.Context, output *TxEssenceOutput) error {
			return nil
		}))
```

So I think it would be worth to see if we can remove `Transaction.syntacticallyValidate` and have serix validators for that instead?

vm/nova/vm.go
cyberphysic4l: Good point. This also needs to be included in the TotalManaIn calculation upon initialisation of the working set, and perhaps elsewhere.
I do think that rewards should be locked to the account that has earned them, but that does not necessarily need to be the case.
We also need to include this in the account TIP.

vm/nova/vm.go
```diff
@@ -512,6 +452,10 @@ func accountStakingSTVF(vmParams *vm.Params, chainID iotago.ChainID, current *io
 	_, isClaiming := vmParams.WorkingSet.Rewards[chainID]
 
 	if currentStakingFeat != nil {
+		if next.FeatureSet().BlockIssuer() == nil {
```
cyberphysic4l: What if the staking feature was removed here? Then it would  be ok for next block issuer feature to be removed too, right? This is both state and governance transition now, so one transaction could remove both block issuer feature and staking feature I think.

output_anchor.go
```diff
@@ -0,0 +1,288 @@
...
+	AnchorOutputFeatures         = Features[anchorOutputFeature]
+	AnchorOutputImmFeatures      = Features[anchorOutputImmFeature]
+)
+
+// AnchorOutput is an output type which represents an anchor.
+type AnchorOutput struct {
+	// The amount of IOTA tokens held by the output.
+	Amount BaseToken `serix:"0,mapKey=amount"`
+	// The stored mana held by the output.
+	Mana Mana `serix:"1,mapKey=mana"`
```
PhilippGackstatter, line 97: I imagine most anchors would typically also have a block issuer account that issues the anchor updates, where they would store most, if not all or their Mana for throughput reasons. Is there a particular reason why we want Anchor Outputs to be able to hold Mana other than "all other outputs can do it"?

I just want to avoid making the same "mistake" with Mana as we did with Native Tokens: First letting all outputs hold it and later restricting it to a few.

output_test.go
```diff
@@ -541,92 +570,156 @@ func TestOutputsSyntacticalAccount(t *testing.T) {
...
-						&iotago.GovernorAddressUnlockCondition{Address: tpkg.RandAccountAddress()},
+						&iotago.AddressUnlockCondition{Address: tpkg.RandAccountAddress()},
 					},
 				},
 			},
 			wantErr: iotago.ErrAccountOutputNonEmptyState,
 		},
 		{
-			name: "fail - cyclic state controller",
+			name: "fail - cyclic",
```
PhilippGackstatter:
```suggestion
			name: "fail - account's unlock condition contains its own account address",
```

vm/nova/vm.go
```diff
@@ -689,6 +668,111 @@ func accountDestructionValid(vmParams *vm.Params, input *vm.ChainOutputWithIDs)
 	return nil
 }
 
+// For output AnchorOutput(s) with non-zeroed AnchorID, there must be a corresponding input AnchorOutput where either its
+// AnchorID is zeroed and StateIndex is zero or an input AnchorOutput with the same AnchorID.
+//
+// On anchor state transitions: The StateIndex must be incremented by 1 and Only Amount, StateIndex and StateMetadata can be mutated.
+//
+// On anchor governance transition: Only StateController (must be mutated), GovernanceController and the MetadataBlock can be mutated.
```
cyberphysic4l: Is it true that the state controller must be mutated? I don't see that enforced.

vm/nova/vm.go
PhilippGackstatter: The rewards are currently added to `Total Mana In` in `ExecFuncBalancedMana` for some reason. This should also be part of the `TotalManaIn` function imo.

The TIPs are correct I'd say. [TIP-42 (BIF)](https://github.com/iotaledger/tips/blob/tip42/tips/TIP-0042/tip-0042.md#additional-semantic-transaction-validation-rule) just uses `Total Mana In` which is basically calculated with the function in [TIP-39](https://github.com/iotaledger/tips/blob/tip39/tips/TIP-0039/tip-0039.md#mana-transaction-validation-rules). Just added a link from TIP-42 to TIP-39 to clarify that.

vm/nova/vm.go
PhilippGackstatter: Yeah, we can't add this check here because this STVF is also called when we have a staking feature on the input side and not on the output side (e.g. when claiming rewards), and in this case there is no BIF required on the output side. 

I think we have these cases where we need the check:
- `accountStakingGenesisValidation` has the check, which is called in both new account and account transition cases.
- `accountStakingExpiredValidation` also calls `accountStakingGenesisValidation` when appropriate, so it also has the check.
- `accountStakingNonExpiredValidation` is where we need to add the check because we could now remove the BIF in the same transaction (due to this now basically being a state transition and governor transition at the same time). Perhaps we can just pass `nextHasBlockIssuerFeat bool` to that function.

vm/nova/stvf_test.go
```diff
@@ -1511,30 +1332,28 @@ func TestAccountOutput_ValidateStateTransition(t *testing.T) {
 			wantErr: nil,
 		},
 		{
-			name: "ok - gov transition",
+			name: "ok - state transition",
```
PhilippGackstatter:
```suggestion
			name: "ok - foundry counter increased by number of new foundries",
```

vm/nova/stvf_test.go
```diff
@@ -1811,343 +1571,660 @@ func TestAccountOutput_ValidateStateTransition(t *testing.T) {
...
+				Amount:    100,
+				AccountID: exampleAccountID,
+				Conditions: iotago.AccountOutputUnlockConditions{
+					&iotago.AddressUnlockCondition{Address: exampleAddress},
+				},
+				Features: iotago.AccountOutputFeatures{
+					&iotago.MetadataFeature{Data: []byte("1337")},
+					&iotago.SenderFeature{Address: exampleAddress},
+					&iotago.BlockIssuerFeature{
+						BlockIssuerKeys: tpkg.RandomBlockIsssuerKeysEd25519(1),
```
PhilippGackstatter:
```suggestion
						BlockIssuerKeys: tpkg.RandomBlockIssuerKeysEd25519(1),
```

:snake:

vm/nova/stvf_test.go
```diff
@@ -1811,343 +1571,660 @@ func TestAccountOutput_ValidateStateTransition(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			wantErr: nil,
+		},
+		{
+			name: "fail - state transition",
```
PhilippGackstatter:
```suggestion
			name: "fail - invalid foundry counters",
```

vm/nova/vm_test.go
```diff
@@ -6072,6 +6768,8 @@ func TestTxSemanticMana(t *testing.T) {
 				wantErr: iotago.ErrManaOverflow,
 			}
 		}(),
+
+		// fail - mana overflow on the output side sum
```
PhilippGackstatter, line 6772: Maybe you should make a VS Code plugin that generates these as inlay hints :D.

vm/nova/vm_test.go
```diff
@@ -6177,23 +6879,24 @@ func TestManaRewardsClaimingStaking(t *testing.T) {
...
+			CreationSlot: currentSlot,
 		},
 		Outputs: iotago.TxEssenceOutputs{
 			&iotago.AccountOutput{
-				Amount:         OneMi * 5,
-				AccountID:      accountIdent.AccountID(),
-				StateIndex:     2,
-				StateMetadata:  nil,
+				Amount:         OneIOTA * 5,
+				Mana:           12043,
```
PhilippGackstatter: Where is this number coming from?

vm/nova/vm_test.go
muXxer: It would be nice if the code is readable without fancy plugins :P

vm/nova/vm_test.go
muXxer: From executing this test :D

PhilippGackstatter: Can you document this :)?

api_v3.go
muXxer: Let's do it in another PR.

output_anchor.go
muXxer: I see where you are coming from, but you can also extend this to NFTs and also to basic outputs. Why would anyone need Mana in those, if everyone can also store it in their account output. So I would say, better leave it as it is, and let the user decide how they will use it or not.

vm/nova/vm.go
cyberphysic4l: Added the rewards to `TotalManaIn` now. 
I propose that we lock rewards to the account that earned them, so we would add keep this block of code that Max added, and we update TIP 42 so that `AccountIn = AccountInPotential + AccountInStored + AccountInRewards`.

muXxer: We decided that we don't want to lock the mana to the account, because it does not contribute to the throughput anyway.

vm/nova/vm_test.go
muXxer: Changed it to being calculated automatically.

vm/vm.go
```diff
@@ -103,6 +103,23 @@ func TotalManaIn(manaDecayProvider *iotago.ManaDecayProvider, storageScoreStruct
...
+	}
+
+	// whether it's valid to claim rewards is checked in the delegation and staking STVFs.
+	for _, reward := range rewards {
+		var err error
+		totalIn, err = safemath.SafeAdd(totalIn, reward)
+		if err != nil {
+			return 0, ierrors.Wrapf(iotago.ErrManaOverflow, "%w", err)
+		}
+	}
```
PhilippGackstatter: Now we have this twice, probably due to the merge.

PhilippGackstatter (approved):

vm/vm.go
muXxer: :ragepepe:

PR #595, opened 2023-11-03
Metadata refactor

### LINKED ISSUE #491: Metadata refactor

muXxer:
- All Metadata features are now `map[string][]byte` instead of `[]byte`
- Removed `StateMetadata` field in favor of new `StateMetadataFeature`

depends on https://github.com/iotaledger/hive.go/pull/600


gen/feat_metadata.tmpl
```diff
@@ -0,0 +1,83 @@
...
+// Code generated by go generate; DO NOT EDIT. Check gen/ directory instead.
+
+import (
+	"bytes"
+
+	"github.com/iotaledger/hive.go/lo"
+	"github.com/iotaledger/hive.go/serializer/v2"
+)
+
+type {{.Name}}Entries map[string][]byte
```
alexsporn: There is no mention whatsoever about ordering of the map. This makes serialization non-deterministic.

alexsporn: This is now also forcing the use of a key everytime. (e.g. `data` as in every changed testcase)

alexsporn: Why is this better than something like IRC-27 or IRC-30? The protocol does not need the map here to function.

muXxer: The new serialization of maps is deterministic in serix. So this wouldn't be a problem anymore :+1:

alexsporn (approved):

PR #596, opened 2023-11-07
Cache V3ProtocolParameter bytes.

### LINKED ISSUE #458: Calculate V3ProtocolParameters Bytes, Hashes and NetworkID only once
In the `V3ProtocolParameters`, we hash every time someone calls `Hash` and `NetworkID`.
Also serix encoding is done everytime `Bytes` is called. We could cache that with sync.Once.

piotrm50:
Cache V3ProtocolParameter serialized bytes NetworkID and hash not to serialize the whole thing each time those methods are called.


alexsporn:
Please check this with the iota-core unit tests with `-race` since I am pretty sure it will complain. We had this before for some of the computed values and it complained.

piotrm50:
> Please check this with the iota-core unit tests with `-race` since I am pretty sure it will complain. We had this before for some of the computed values and it complained.

I remember the problem. I checked and it doesn't complain. I think before it complained exactly because serix was reading the values without looking at mutexes, but now after the first serialization we don't serialize again. And when we do serialize, then those cache fields are not modified.

alexsporn:
> > Please check this with the iota-core unit tests with `-race` since I am pretty sure it will complain. We had this before for some of the computed values and it complained.
> 
> I remember the problem. I checked and it doesn't complain. I think before it complained exactly because serix was reading the values without looking at mutexes, but now after the first serialization we don't serialize again. And when we do serialize, then those cache fields are not modified.

Then we can maybe move some of the computed values from the API back to the ProtocolParameters

muXxer (approved):

PR #597, opened 2023-11-07
Unify syntactic output checks & make address restriction checks syntactic

### LINKED ISSUE #471: Make Address Capability checks syntactic
They are currently semantic but since only the outputs are required to execute that check, they can be made syntactic instead.

PhilippGackstatter:
# Description of change

Unifies the syntactic checks for outputs. Some were executed at (de)serialization time, some were run as part of syntactic transaction validation. Now they are all run in the latter since that is the more established pattern.

Also makes the address restriction checks syntactic (rather than semantic) and introduces a dedicated function for the semantic check that only one input in a given transaction may contain an address of type implicit account creation address.

Finally, this PR makes `SyntacticallyValidate` on `TransactionEssence` private, renames it and makes `syntacticallyValidate` on the `Transaction` public. Since the essence is embedded in the transaction it was easy to accidentally call `transaction.SyntacticallyValidate` which would only validate the inner essence instead of the whole transaction. That was already the case in the signed transaction where `t.Transaction.syntacticallyValidate(t.API)` was called.

Some tests had to be adapted because they relied on serix validation. The previous deserialization tests for implicit account creation address checks were moved to a dedicated test.

### DIFF
--- a/output_test.go
+++ b/output_test.go
@@ -29,8 +29,6 @@ func TestOutputTypeString(t *testing.T) {
 }
 
 func TestOutputsDeSerialize(t *testing.T) {
-	emptyAccountAddress := iotago.AccountAddress(iotago.EmptyAccountID)
-
 	tests := []deSerializeTest{
 		{
 			name: "ok - BasicOutput",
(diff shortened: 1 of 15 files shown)


output_test.go
```diff
@@ -939,7 +868,7 @@ func TestOutputsSyntacticaDelegation(t *testing.T) {
 					},
 				},
 			},
-			wantErr: iotago.ErrDelegationValidatorAddressZeroed,
+			wantErr: iotago.ErrDelegationValidatorAddressEmpty,
```
muXxer, line 871: can you also change the name of that test? that "zeroed" is confusing :D

muXxer:
```
fail - Delegation Output contains empty validator address
```

muXxer (approved):

PR #600, opened 2023-11-08
Adapt to serix changes

muXxer:


vm/types.go
```diff
@@ -73,7 +73,7 @@ type ResolvedInputs struct {
 }
 
 type ImplicitAccountOutput struct {
-	*iotago.BasicOutput `serix:"0,nest=false"`
+	*iotago.BasicOutput `serix:""`
```
alexsporn: does this need a serix tag?

alexsporn (approved):

vm/types.go
muXxer: Removed

PR #604, opened 2023-11-10
Allow Features in Implicit Accounts

PhilippGackstatter:
# Description of change

Allow features in implicit accounts because they allow for more seamless implicit account creation. The more permutations of Basic Outputs we allow implicit account creation addresses in, the more seamless implicit account creation becomes.

On the other hand, Storage Deposit Return, Timelock and Expiration Unlock Condition are not useful for implicit account creation, so they continue to be disallowed.

- Storage Deposit Return UC: Having to return the funds on an implicit account to the sender means the full account we want to transition to must be funded by other means. In a typical onboarding scenario these funds do not exist, so SDRUC would only make it more complicated to create the full account.
- Timelock: Would be harmless, but timelocks are not permitted on full accounts either. It would be odd to be able to use the implicit account to issue blocks (because the accounts ledger does not respect the timelock) but not be able to do the implicit account conversion until the timelock expires.
- Expiration: Necessarily, the the `Return Address` would not be an implicit account creation address, but probably a regualr `Ed25519 Address`. Then for the `Address` the Basic Output would be an implicit account, but for the `Return Address` it would be a regular output, which is very undesirable behavior.

Added new tests for where implicit account creation addresses are allowed.


vm/nova/vm_test.go
```diff
@@ -7444,36 +7447,86 @@ func TestTxSemanticImplicitAccountCreationAndTransition(t *testing.T) {
...
+			name:   "ok - implicit account contains features",
+			inputs: exampleInputs,
+			outputs: []iotago.Output{
+				&iotago.BasicOutput{
+					Amount: exampleAmount,
+					UnlockConditions: iotago.BasicOutputUnlockConditions{
+						&iotago.AddressUnlockCondition{Address: implicitAccountIdent},
+					},
+					Features: iotago.BasicOutputFeatures{
+						&iotago.MetadataFeature{
```
muXxer: Don't forget to update your branch, otherwise this will clash with #595

vm/nova/vm_test.go
```diff
@@ -7408,6 +7408,9 @@ func TestTxSemanticImplicitAccountCreationAndTransition(t *testing.T) {
 			&iotago.AddressUnlockCondition{Address: implicitAccountIdent},
 		},
 	}
+	exampleMetadataFeature := iotago.MetadataFeature{Data: tpkg.RandBytes(40)}
```
muXxer: why not a pointer here like the other features in the tests?

muXxer (approved):

PR #606, opened 2023-11-14
Protocol params sanity checks

### LINKED ISSUE #512: Sanity checks for protocol parameters & derive parameters as much as possible
It would be great to have sanity checks for protocol parameters in place that evaluate whether a certain combination of parameters makes sense. This could, for example happen, when constructing the protocol parameters and the function that does all these checks could serve as a documentation going forward as to which parameters are related and how. 

The functionality should have different levels of escalation:
- for sensible parameters for a live network: this should warn, so that it's still possible to overwrite them e.g. in unit or end-to-end tests
- impossible combinations of parameters: this should panic

cyberphysic4l:
Add sanity checks to protocol parameters where we have certain constraints e.g. MaxCommittableAge >= MinCommittableAge

closes https://github.com/iotaledger/iota-core/issues/512


karimodm:
We are recently having discussions about deriving as many protocol parameters as possible to partially avoid these sanity checks. Would this be the right PR to start deriving some?

cyberphysic4l:
> We are recently having discussions about deriving as many protocol parameters as possible to partially avoid these sanity checks. Would this be the right PR to start deriving some?

Yes, I think that would be a good task to combine with deriving some. Maybe let's start making a list of what we might like to make derived in the issue? https://github.com/iotaledger/iota-core/issues/512

oliviasaa:
can we add these checks as well?
https://github.com/iotaledger/iota-core/issues/450

oliviasaa:
what is missing is:
- decayFactorEpochsSum*generationRate must use at most 32 bits
- 21 * S * gamma * d/(beta*Delta) must use at most bitsCount bits

where 
- S is the token supply, 
- gamma is the generation rate per slot, meaning generationRate*2^(-generationRateExp)
- d is the number of slots per epoch
- beta is the decay factor per year (1/3)
- Delta is the epoch duration in years

This is something that cannot be "derived"; those are consistency checks to prevent overflowing in some functions.

oliviasaa:
I'm having a problem testing this repo locally and I don't want to break anything. I created another branch so you can see what I did.

api_v3_protocol_parameters.go
```diff
@@ -310,17 +349,28 @@ func WithManaOptions(bitsCount uint8, generationRate uint8, generationRateExpone
...
 
-func WithTimeProviderOptions(genesisSlot SlotIndex, genesisTimestamp int64, slotDurationInSeconds uint8, slotsPerEpochExponent uint8) options.Option[V3ProtocolParameters] {
+func WithTimeOptions(genesisSlot SlotIndex, genesisTimestamp int64, slotDurationInSeconds uint8, slotsPerEpochExponent uint8, livenessThresholdLowerBoundInSeconds uint16, livenessThresholdUpperBoundInSeconds uint16, minCommittableAge SlotIndex, maxCommittableAge SlotIndex, epochNearingThreshold SlotIndex) options.Option[V3ProtocolParameters] {
+	if livenessThresholdLowerBoundInSeconds > livenessThresholdUpperBoundInSeconds {
+		panic("livenessThresholdLowerBoundInSeconds must be smaller than livenessThresholdUpperBoundInSeconds")
+	}
+	if SlotIndex(livenessThresholdUpperBoundInSeconds) > minCommittableAge*SlotIndex(slotDurationInSeconds) {
+		panic("livenessThresholdUpperBoundInSeconds * slotDurationInSeconds must be smaller than minCommittableAge")
+	}
+	if minCommittableAge > maxCommittableAge {
```
bingyanglin: >=

api_v3_protocol_parameters.go
```diff
@@ -310,17 +349,28 @@ func WithManaOptions(bitsCount uint8, generationRate uint8, generationRateExpone
 	}
 }
 
-func WithTimeProviderOptions(genesisSlot SlotIndex, genesisTimestamp int64, slotDurationInSeconds uint8, slotsPerEpochExponent uint8) options.Option[V3ProtocolParameters] {
+func WithTimeOptions(genesisSlot SlotIndex, genesisTimestamp int64, slotDurationInSeconds uint8, slotsPerEpochExponent uint8, livenessThresholdLowerBoundInSeconds uint16, livenessThresholdUpperBoundInSeconds uint16, minCommittableAge SlotIndex, maxCommittableAge SlotIndex, epochNearingThreshold SlotIndex) options.Option[V3ProtocolParameters] {
+	if livenessThresholdLowerBoundInSeconds > livenessThresholdUpperBoundInSeconds {
+		panic("livenessThresholdLowerBoundInSeconds must be smaller than livenessThresholdUpperBoundInSeconds")
+	}
+	if SlotIndex(livenessThresholdUpperBoundInSeconds) > minCommittableAge*SlotIndex(slotDurationInSeconds) {
```
bingyanglin: >=

oliviasaa:
TO DO:

- [x] derive bootstrapping duration from beta
- [x] review sanity checks for decay factor sum and other mana related parameters

mana_decay_provider.go
```diff
@@ -244,35 +237,46 @@ func (p *ManaDecayProvider) GenerateManaAndDecayBySlots(amount BaseToken, creati
 		if err != nil {
 			return 0, ierrors.Wrap(err, "failed to decay generated mana in the first epoch")
 		}
+		/*
+			manaGeneratedWholeEpoch, err := p.generateMana(amount, 1<<p.slotsPerEpochExponent)
```
muXxer: what is that here?

cyberphysic4l: Don't know, some workings from Olivia... Deleted.

muXxer (approved):

PR #611, opened 2023-11-21
Added new wallet pkg with KeyManager and Account helpers from the iota-core mock package

alexsporn:
In this package we can implement common wallet operations


go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/pasztorpisti/qs v0.0.0-20171216220353-8d6c33ee906c
 	github.com/samber/lo v1.38.1
 	github.com/stretchr/testify v1.8.4
+	github.com/wollac/iota-crypto-demo v0.0.0-20221117162917-b10619eccb98
```
muXxer, line 19: maybe we should use the fork?
should we rename the package in the fork so we don't need a replace statement?

wallet/account.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+// ID returns the accountID.
+func (e *Ed25519Account) ID() iotago.AccountID {
+	return e.accountID
+}
+
+// Address returns the account address.
+func (e *Ed25519Account) Address() iotago.Address {
```
muXxer: That confused me a bit.
It's not the account address derived from the accountID, but the Ed25519 address derived from the private key. :sweat_smile: 
Can we change the function name and/or the comment?

wallet/account.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return iotago.Ed25519AddressFromPubKey(ed25519PubKey)
+}
+
+// PrivateKey returns the account private key for signing.
+func (e *Ed25519Account) PrivateKey() ed25519.PrivateKey {
+	return e.privateKey
+}
+
+func AccountFromParams(accountHex string, privateKey string) Account {
```
muXxer: Maybe return the errors instead of panic?

wallet/account.go
```diff
@@ -0,0 +1,71 @@
...
+	iotago "github.com/iotaledger/iota.go/v4"
+)
+
+// Account represents an account.
+type Account interface {
+	// ID returns the accountID.
+	ID() iotago.AccountID
+
+	// Address returns the account address.
+	Address() iotago.Address
```
muXxer: See my comment in the implementation.

wallet/keymanager.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/wollac/iota-crypto-demo/pkg/bip32path"
+	"github.com/wollac/iota-crypto-demo/pkg/slip10"
+	"github.com/wollac/iota-crypto-demo/pkg/slip10/eddsa"
+
+	"github.com/iotaledger/hive.go/ierrors"
+	iotago "github.com/iotaledger/iota.go/v4"
+)
+
+const (
+	pathString = "44'/4218'/0'/%d'"
```
muXxer: Maybe this should be a parameter to the keymanager instead.

muXxer (changes requested):

wallet/account.go
alexsporn: That is why I told you to wait with the review, exactly because I was refactoring this, hehe

wallet/account.go
```diff
@@ -56,16 +62,16 @@ func (e *Ed25519Account) PrivateKey() ed25519.PrivateKey {
 	return e.privateKey
 }
 
-func AccountFromParams(accountHex string, privateKey string) Account {
+func AccountFromParams(accountHex string, privateKey string) (Account, error) {
 	accountID, err := iotago.AccountIDFromHexString(accountHex)
 	if err != nil {
-		panic(fmt.Sprintln("invalid accountID hex string", err))
+		return nil, ierrors.Wrapf(err, "invalid accountID hex string")
```
muXxer: Wrap

wallet/account.go
```diff
@@ -56,16 +62,16 @@ func (e *Ed25519Account) PrivateKey() ed25519.PrivateKey {
...
 	}
 
 	privKey, err := crypto.ParseEd25519PrivateKeyFromString(privateKey)
 	if err != nil {
-		panic(fmt.Sprintln("invalid ed25519 private key string", err))
+		return nil, ierrors.Wrapf(err, "invalid ed25519 private key string")
```
muXxer: Wrap

wallet/account.go
```diff
@@ -41,8 +44,11 @@ func (e *Ed25519Account) ID() iotago.AccountID {
 	return e.accountID
 }
 
-// Address returns the account address.
-func (e *Ed25519Account) Address() iotago.Address {
+func (e *Ed25519Account) Address() *iotago.AccountAddress {
+	return e.accountID.ToAddress().(*iotago.AccountAddress)
```
muXxer, line 49: nolint :D

go.mod
alexsporn: Lets do that later. We need to decide where to move the code to

muXxer (approved):

PR #613, opened 2023-11-22
fix: make address.NewAddress public

jorgemmsilva:
minor change

### DIFF
--- a/address.go
+++ b/address.go
@@ -174,7 +174,7 @@ type UTXOIDChainID interface {
 	FromOutputID(id OutputID) ChainID
 }
 
-func newAddress(addressType AddressType) (address Address, err error) {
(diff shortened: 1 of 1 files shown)


alexsporn:
I think you are using it wrong. You should not need to initialize an empty address directly.
To decode an address from bytes you should do:
```
var addr iotago.Address
byteCount, err := iotago.CommonSerixAPI().Decode(context.Background(), bytes, &addr)
```
If needed we can wrap this in an `AddressFromBytes` method.

To serialize you can do `addr.Bytes()`.

jorgemmsilva:
thanks @alexsporn I think you're right.
I was blindingly following this comment https://github.com/iotaledger/wasp/blob/iota2.0/packages/isc/address.go#L13
closing :+1:

PR #620, opened 2023-11-23
Add Transaction Metadata endpoint and response structure

jkrvivian:


muXxer:
What about the method in the `http_api_client.go`?

muXxer (approved):

PR #624, opened 2023-11-27
Fix BIC check when Account ID is zeroed

PhilippGackstatter:
# Description of change

Fix BIC check when Account ID is zeroed by always using the pre-calculated chain ID of the input rather than the `AccountID` field of the input which may be zero.

## Type of change

Adapted tests and added a new one so both modified code paths are covered.

### DIFF
--- a/vm/nova/vm.go
+++ b/vm/nova/vm.go
@@ -298,7 +298,12 @@ func accountStateChangeValid(vmParams *vm.Params, input *vm.ChainOutputWithIDs, 
 	// If a Block Issuer Feature is present on the input side of the transaction,
 	// and the BIC is negative, the account is locked.
 	if current.FeatureSet().BlockIssuer() != nil {
-		if bic, exists := vmParams.WorkingSet.BIC[current.AccountID]; exists {
+		accountID, is := input.ChainID.(iotago.AccountID)
+		if !is {
+			return ierrors.Wrapf(iotago.ErrBlockIssuanceCreditInputRequired, "cannot convert chain ID to account ID")
+		}
+
+		if bic, exists := vmParams.WorkingSet.BIC[accountID]; exists {
 			if bic < 0 {
 				return ierrors.Wrap(iotago.ErrAccountLocked, "negative block issuer credit")
 			}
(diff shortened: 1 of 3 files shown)


vm/nova/vm.go
```diff
@@ -298,7 +298,12 @@ func accountStateChangeValid(vmParams *vm.Params, input *vm.ChainOutputWithIDs,
 	// If a Block Issuer Feature is present on the input side of the transaction,
 	// and the BIC is negative, the account is locked.
 	if current.FeatureSet().BlockIssuer() != nil {
-		if bic, exists := vmParams.WorkingSet.BIC[current.AccountID]; exists {
+		accountID, is := input.ChainID.(iotago.AccountID)
+		if !is {
+			return ierrors.Wrapf(iotago.ErrBlockIssuanceCreditInputRequired, "cannot convert chain ID to account ID")
```
muXxer: Either add the `ChainID.ToHex()` or only do a `Wrap` (I prefer the first)

vm/nova/vm.go
```diff
@@ -623,7 +628,12 @@ func accountDestructionValid(vmParams *vm.Params, input *vm.ChainOutputWithIDs)
...
-		if bic, exists := vmParams.WorkingSet.BIC[outputToDestroy.AccountID]; exists {
+		accountID, is := input.ChainID.(iotago.AccountID)
+		if !is {
+			return ierrors.Wrapf(iotago.ErrBlockIssuanceCreditInputRequired, "cannot convert chain ID to account ID")
+		}
+
+		if bic, exists := vmParams.WorkingSet.BIC[accountID]; exists {
 			if bic < 0 {
 				return ierrors.Wrap(iotago.ErrAccountLocked, "cannot destroy locked account")
 			}
```
muXxer: Same here, should we add a helper function for this btw?

PhilippGackstatter: Not sure what helper function you actually mean, but I think it's fine to just include the chain id in the error. Should never happen anyway because we're already in the account code path. If it fails, it's a bug in the VM.

muXxer (approved):

PR #625, opened 2023-11-28
Cleanup API and field order

### LINKED ISSUE #623: Order of BlockIssuerFeature's fields
The fields of `BlockIssuerFeature` are not in the same order as in the TIP, leading to a different serialization than the SDK and then a failure to parse sent blocks.

iota.go: https://github.com/iotaledger/iota.go/blob/develop/feat_blockissuer.go#L16-L19
TIP: https://github.com/iotaledger/tips/blob/tip42/tips/TIP-0042/tip-0042.md#block-issuer-feature

muXxer:
There was a wrong order in the blockissuer feature which cause the serialization not to be aligned with the TIPs.

The PR also cleans up the CORE API to have a consistent naming scheme.
Some responses were also simplified like the output metadata response.

### DIFF
--- a/api/core.go
+++ b/api/core.go
@@ -248,8 +248,8 @@ type (
 		ShallowLikeParents iotago.BlockIDs `serix:",lenPrefix=uint8,omitempty"`
 		// LatestFinalizedSlot is the latest finalized slot.
 		LatestFinalizedSlot iotago.SlotIndex `serix:""`
-		// Commitment is the latest commitment of the node.
-		Commitment *iotago.Commitment `serix:""`
+		// LatestCommitment is the latest commitment of the node.
+		LatestCommitment *iotago.Commitment `serix:""`
 	}
 
 	// BlockCreatedResponse defines the response of a POST blocks REST API call.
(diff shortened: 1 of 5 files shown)


api/core.go
```diff
@@ -294,22 +292,34 @@ type (
 		OutputIDProof *iotago.OutputIDProof  `serix:""`
 	}
 
+	OutputInclusionMetadata struct {
+		// Slot is the slot in which the output was included.
+		Slot iotago.SlotIndex `serix:""`
+		// TransactionID is the transaction ID that creates the output.
```
PhilippGackstatter:
```suggestion
		// TransactionID is the transaction ID that created the output.
```

api/core.go
```diff
@@ -294,22 +292,34 @@ type (
...
+		TransactionID iotago.TransactionID `serix:""`
+		// CommitmentID is the commitment ID that includes the creation of the output.
+		CommitmentID iotago.CommitmentID `serix:",omitempty"`
+	}
+
+	OutputConsumptionMetadata struct {
+		// Slot is the slot in which the output was spent.
+		Slot iotago.SlotIndex `serix:""`
+		// TransactionID is the transaction ID that spends the output.
```
PhilippGackstatter:
```suggestion
		// TransactionID is the transaction ID that spent the output.
```

PhilippGackstatter (approved):

PR #628, opened 2023-11-30
Extend tpkg funcs

muXxer:


tpkg/util.go
```diff
@@ -36,44 +41,70 @@ func RandByte() byte {
...
 
 	return b
 }
 
 func RandString(length int) string {
-	return string(RandBytes(length))
+	var b []byte
+	for i := 0; i < length; i++ {
+		// 128 because we want to have printable ASCII characters
+		b = append(b, byte(RandInt(128)))
```
PhilippGackstatter: If we want to have it printable we should limit it to `32` to `126`, because you don't want the control characters in a string. Or why not just generate a random alphanumeric string?

tpkg/util.go
```diff
@@ -734,6 +796,17 @@ func RandBlock(blockBody iotago.BlockBody, api iotago.API, rmc iotago.Mana) *iot
 	}
 }
 
+func RandBasicBlockWithPayload(api iotago.API, payload iotago.ApplicationPayload) *iotago.BasicBlockBody {
```
PhilippGackstatter: Maybe we should call this one (and friends) `RandBasicBlockBody...` because they don't return an actual Basic Block (which would be a `iotago.Block` with a `BasicBlockBody`).

PhilippGackstatter (approved):

PR #641, opened 2023-12-08
Rewards sanity check

### LINKED ISSUE #614: Panic on rewards calculation overflow
While running committee rotation tests in DockerTestFramework ( https://github.com/iotaledger/iota-core/pull/600), nodes panic because the rewards calculation overflow.

To reproduce the errors, checkout to  https://github.com/iotaledger/iota-core/pull/600, and run the `Test_SmallerCommittee` in `tools/docker-network/tests/committee_rotation.go`. With `go.work`, every library is the latest version.

Here's the protocol parameters setup:

```
committeeTargetSize = 4
slotsPerEpochExponent = 4
livenessThresholdUpperBound = 10
livenessThresholdLowerBound = 10
minCommittableAge = 2
maxCommittableAge = 4
epochNearingThreshold=8
```

And the logs:
```
2023-12-07T06:20:56Z	DEBUG	Protocol	app/component.go:98	ConfirmedTimeUpdated, slot: 38 @ 2023-12-07 06:20:55.090446205 +0000 UTC
panic: integer under- or overflow: 2476192049932568352 * 10: failed to calculate pool reward without fixed costs due to overflow for slot 36: failed to calculate pool rewards for account 0x6aee704f25558e8aa7630fed0121da53074188abc423b3c5810f80be4936eb6e

goroutine 381 [running]:
github.com/iotaledger/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1/performance.(*Tracker).ApplyEpoch.func1({0x6a, 0xee, 0x70, 0x4f, 0x25, 0x55, 0x8e, 0x8a, 0xa7, 0x63, ...}, ...)
	/scratch/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go:253 +0x3f3
github.com/iotaledger/hive.go/ds/shrinkingmap.(*ShrinkingMap[...]).ForEach(0x2e0d9c0, 0xc000ebd618)
	/scratch/hive.go/ds/shrinkingmap/shrinkingmap.go:174 +0x275
github.com/iotaledger/iota-core/pkg/core/account.(*Accounts).ForEach(...)
	/scratch/iota-core/pkg/core/account/accounts.go:125
github.com/iotaledger/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1/performance.(*Tracker).ApplyEpoch(0xc0004f8500, 0x1, 0xc00142da00)
	/scratch/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1/performance/performance.go:214 +0x369
github.com/iotaledger/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1.(*SybilProtection).CommitSlot(0xc000634bb0, 0x24)
	/scratch/iota-core/pkg/protocol/sybilprotection/sybilprotectionv1/sybilprotection.go:190 +0x2d2
github.com/iotaledger/iota-core/pkg/protocol/engine/notarization/slotnotarization.(*Manager).createCommitment(0xc0003760f0, 0x24)
	/scratch/iota-core/pkg/protocol/engine/notarization/slotnotarization/manager.go:227 +0x559
github.com/iotaledger/iota-core/pkg/protocol/engine/notarization/slotnotarization.(*Manager).tryCommitSlotUntil(0xc0003760f0, 0x26)
	/scratch/iota-core/pkg/protocol/engine/notarization/slotnotarization/manager.go:191 +0x92
github.com/iotaledger/iota-core/pkg/protocol/engine/notarization/slotnotarization.(*Manager).tryCommitUntil(0xc0003760f0, 0x26)
	/scratch/iota-core/pkg/protocol/engine/notarization/slotnotarization/manager.go:120 +0x45
github.com/iotaledger/iota-core/components/protocol.provide.func1.NewProvider.func14.1.1(0xc000ebded8?)
	/scratch/iota-core/pkg/protocol/engine/notarization/slotnotarization/manager.go:78 +0x131
github.com/iotaledger/hive.go/runtime/event.(*Event1[...]).Trigger.func1.1()
	/scratch/hive.go/runtime/event/events.go:86 +0x1f
github.com/iotaledger/hive.go/runtime/workerpool.(*Task).run(0xc001a9f080)
	/scratch/hive.go/runtime/workerpool/task.go:48 +0x72
github.com/iotaledger/hive.go/runtime/workerpool.(*WorkerPool).workerReadLoop(0xc00028ec40)
	/scratch/hive.go/runtime/workerpool/workerpool.go:190 +0x25
github.com/iotaledger/hive.go/runtime/workerpool.(*WorkerPool).worker(0xc00028ec40)
	/scratch/hive.go/runtime/workerpool/workerpool.go:170 +0x50
created by github.com/iotaledger/hive.go/runtime/workerpool.(*WorkerPool).startWorkers in goroutine 354
	/scratch/hive.go/runtime/workerpool/workerpool.go:162 +0x25

```

cyberphysic4l:
Adds a sanity check that token supply bits count + poolCoefficientExponent is less that 64 to fit in the int64 range during rewards calculations.


api_v3_protocol_parameters.go
```diff
@@ -349,6 +350,12 @@ func congestionControlSanityCheck(protocolParams *V3ProtocolParameters) {
 	}
 }
 
+func rewardsSanityCheck(protocolParams *V3ProtocolParameters) {
+	if math.Log2(float64(protocolParams.TokenSupply()))+float64(protocolParams.RewardsParameters().PoolCoefficientExponent) > 63 {
+		panic("Token supply bits count + poolCoefficientExponent must be less than or equal to 63")
+	}
```
PhilippGackstatter:
```suggestion
    tokenSupplyBitsCount := uint8(math.Log2(float64(protocolParams.TokenSupply())))
	if tokenSupplyBitsCount + protocolParams.RewardsParameters().PoolCoefficientExponent > 63 {
		message := fmt.Sprintf("Token supply bits count (%d) + PoolCoefficientExponent (%d) must be less than or equal to 63\n", tokenSupplyBitsCount, protocolParams.RewardsParameters().PoolCoefficientExponent)
		panic(message)
	}
```

PhilippGackstatter (commented):
Small improvement to use integers for the comparison rather than floats and print the actual values. We do this in the congestion control sanity check and it was already quite helpful when I set custom parameters..

Also would you mind adding another sanity check to ensure `ValidationBlocksPerSlot <= 32`? We currently have that [assumption](https://github.com/iotaledger/iota-core/blob/a3779975dcf052507e7ffa425dc2e841192e07ba/pkg/model/validator_performance.go#L11) but it's not checked anywhere afaik, so in the protocol params would be good.

Edit: I think it's `<= 32` because `ValidationBlocksPerSlot` equals the number of subslots and a `uint32` can track 32 bits (=subslots) so it's fine to be equal to 32.

api_v3_protocol_parameters.go
PhilippGackstatter: Maybe @oliviasaa can comment on whether it should be `> 63` or `>= 63`. I'm not sure right now.

Also I guess if we truncate the float result to integer it would have to be `uint8(log2(tokenSupply)) + 1` in case there is a decimal place?

cyberphysic4l: If you use uint8 in this way, then you also need to check for overflows incase someone were to set PoolCoefficentExponent = 255 for example, then the sanity check would pass but this would obviously not be a sane value.

cyberphysic4l: I am actually also sceptical about one of the other assumptions in the comments/TIP, which is that the target reward is at most 42 bits. Where does this assumption come from and does it also need to be sanity checked?

cyberphysic4l: Regarding the >63 or >=63 point, I actually think the check can be >64 if, as you suggested, we use the actual bit count used by the token supply which is given by uint8(log2(tokenSupply))+1. So the pool stake takes up tokenSupplyBits, and that gets shifted left by poolCoefficientExponent, so that can be up to 64 bits without issue. Still happy to wait for @oliviasaa to check though.

PhilippGackstatter (approved):
Looks good to me, thanks for the Validation Blocks check!

api_v3_protocol_parameters.go
oliviasaa: <64 :)

oliviasaa: and I agree with Andrew that using floats for the tests is simpler, because of a potential overflowing

oliviasaa: >that the target reward is at most 42 bits

this was in the initial calculations, which used the iota supply, but you can just check if Computed Initial Rewards uses this amount of bits

oliviasaa (approved):

api_v3_protocol_parameters.go
cyberphysic4l: > this was in the initial calculations, which used the iota supply, but you can just check if Computed Initial Rewards uses this amount of bits

@oliviasaa are you saying that the reference to 42 bits is outdated and should be updated to something else? Do we need another check that Computed Initial Rewards uses some amount of bits? The tip should be updated and so should the comment in the code to make this understandable either way.

muXxer (approved):

PhilippGackstatter:
Let's get this merged to fix iotaledger/iota-core#614 and address potential other checks in future PRs.

PR #645, opened 2023-12-13
Improve determinig the staking reward claiming condition

PhilippGackstatter:
# Description of change

Improve determinig the staking claiming condition. Claiming rewards is "deny by default" in the sense that unless a bool flag is set to true, rewards may not be claimed. This is a more resilient and easy to understand way than how it was before where rewards were allowed to be added and every code path needed to allow or disallow claiming rewards.

## Type of change

Added more tests.


cyberphysic4l (approved):
The bool pointer reads kinda funny but logic seems good to me

PR #659, opened 2024-01-17
Move syntax checks from serix to iotago

### LINKED ISSUE #654: Sort Metadata and Allotment Entries lexicographically by key
Sort Metadata Entries lexicographically by key to ensure deterministic serialization of transactions.

For the binary format, this is already done, but not for JSON.

I also noticed that Allotments, when serialized as JSON, are not sorted automatically, which is (currently) expected behavior of serix.

PhilippGackstatter:
# Description of change

Implement the syntactical checks for objects with serix validators instead of relying on serix's `ValidationMode`.

The primary reason is that serix's two serialization modes, binary and JSON, are not on par with the syntactical checks. Moreover, at some points we relied on the binary format to execute syntactical checks (allotment lexical order, block max size) which cannot be done for JSON. Hence it makes more sense to use serix only for serialization and the most basic checks (e.g. array min/max rule) and check the invariants of objects on the object itself rather than on its binary representation, which is much less error prone. This is also a good step for a later refactor of serix to use code generation.

Lexical Order and Uniqueness in a slice of objects are checked based on the `Compare` implementation of the object in question. This means we can nicely customize what "uniqueness" means. For unlock conditions and features, we define uniqueness based on the type of unlock condition/feature, which results in the semantics of `AtMostOneOfEachTypeByte` plus uniqueness and gives us lexical ordering, too. This approach therefore makes certain previous checks that iterated over the entire slice of objects unnecessary.

The new functionality this PR introduces is lexical ordering for Context Inputs. Like for any array of objects, lexical order makes it much more efficient to check for uniqueness as no extra memory needs to be allocated.
An interesting case is `CommitmentInput` where we, similar to before, also define uniqueness and lexical order based only on its `ContextInputType`, which immediately gives us "At Most One Commitment Input" semantics, so we can skip the additional check we previously had.

As for the original issue that triggered this PR, the decision was to enforce lexical order of maps for the binary format, but not for JSON.
For binary, the ordering happens in serix itself when encoding. When decoding, the order is checked as well.

## Type of change

Loads of tests.


block.go
```diff
@@ -281,6 +285,16 @@ func (b *Block) syntacticallyValidate() error {
 		return ierrors.Wrapf(ErrBlockNetworkIDInvalid, "got %v, want %v (%s)", b.Header.NetworkID, expectedNetworkID, b.API.ProtocolParameters().NetworkName())
 	}
 
+	if err := SliceValidator(b.Body.StrongParentIDs(), LexicalOrderAndUniquenessValidator[BlockID]()); err != nil {
+		return err
```
muXxer: Maybe return wrapped error messages, so we know which slice was the reason for the error?

transaction_test.go
```diff
@@ -396,3 +401,750 @@ func TestTransactionSyntacticMaxMana(t *testing.T) {
...
+			},
+			seriErr:   iotago.ErrArrayValidationOrderViolatesLexicalOrder,
+			deseriErr: iotago.ErrArrayValidationOrderViolatesLexicalOrder,
+		},
+		{
+			name: "fail - AnchorOutput contains lexically unordered unlock conditions",
+			output: &iotago.AnchorOutput{
+				Amount: 10_000_000,
+				UnlockConditions: iotago.AnchorOutputUnlockConditions{
+					govUnlockCond, stateCtrlUnlockCond,
```
muXxer: why in one line :P

transaction_test.go
```diff
@@ -396,3 +401,750 @@ func TestTransactionSyntacticMaxMana(t *testing.T) {
...
+			},
+			seriErr:   iotago.ErrArrayValidationViolatesUniqueness,
+			deseriErr: iotago.ErrArrayValidationViolatesUniqueness,
+		},
+		{
+			name: "fail - AnchorOutput contains duplicate unlock conditions",
+			output: &iotago.AnchorOutput{
+				Amount: 10_000_000,
+				UnlockConditions: iotago.AnchorOutputUnlockConditions{
+					stateCtrlUnlockCond, stateCtrlUnlockCond, govUnlockCond,
```
muXxer: why in one line :P

transaction_test.go
```diff
@@ -396,3 +401,750 @@ func TestTransactionSyntacticMaxMana(t *testing.T) {
...
+					timelockUnlockCond2,
+				},
+			},
+			seriErr:   iotago.ErrArrayValidationViolatesUniqueness,
+			deseriErr: iotago.ErrArrayValidationViolatesUniqueness,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, test.ToDeserializeTest().deSerialize)
```
muXxer, line 808: what about account, foundry and delegation outputs? Shouldn't we check uniqueness there?

transaction_test.go
```diff
@@ -396,3 +401,750 @@ func TestTransactionSyntacticMaxMana(t *testing.T) {
...
+		source:    tx,
+		target:    &iotago.SignedTransaction{},
+		seriErr:   test.seriErr,
+		deSeriErr: test.deseriErr,
+	}
+}
+
+// Tests that lexical order & uniqueness are checked for unlock conditions across all relevant outputs.
+func TestTransactionOutputUnlockConditionsLexicalOrderAndUniqueness(t *testing.T) {
+	// Unlock Cond Type 0
```
muXxer: Maybe instead of writing the types here, write them next to the line where you actually use them?
It's easier to read I guess.

transaction_test.go
```diff
@@ -396,3 +401,750 @@ func TestTransactionSyntacticMaxMana(t *testing.T) {
...
+					issuerFeat, issuerFeat2,
+				},
+			},
+			seriErr:   iotago.ErrArrayValidationViolatesUniqueness,
+			deseriErr: iotago.ErrArrayValidationViolatesUniqueness,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, test.ToDeserializeTest().deSerialize)
```
muXxer, line 1169: What about the foundry output?
Can you also flip NFT and anchor? My OCD hits hard!

api_test.go
```diff
@@ -24,39 +24,73 @@ type deSerializeTest struct {
...
+		require.ErrorIs(t, err, test.seriErr, "JSON encoding")
+
+		// Encode again without validation so we can check that deserialization would also fail.
+		sourceJSON, err = tpkg.ZeroCostTestAPI.JSONEncode(test.source)
+		require.NoError(t, err, "JSON encoding")
+	} else {
+		require.NoError(t, err, "JSON encoding")
+	}
+
+	// fmt.Println(string(lo.PanicOnErr(tpkg.ZeroCostTestAPI.JSONEncode(test.source))))
```
muXxer: leftover?

block_test.go
```diff
@@ -41,6 +45,66 @@ func TestBlock_DeSerialize(t *testing.T) {
...
+					blockID3,
+				}
+				basicBlockBody.WeakParents = iotago.BlockIDs{}
+
+				return block
+			}(),
+			target: &iotago.Block{},
+		},
+		{
+			name: "ok - basic block strong parent ids unsorted",
```
muXxer: So that's "ok" for you? :trollface:

block_test.go
```diff
@@ -41,6 +45,66 @@ func TestBlock_DeSerialize(t *testing.T) {
...
+				basicBlockBody.WeakParents = iotago.BlockIDs{}
+
+				return block
+			}(),
+			target:    &iotago.Block{},
+			seriErr:   iotago.ErrArrayValidationOrderViolatesLexicalOrder,
+			deSeriErr: iotago.ErrArrayValidationOrderViolatesLexicalOrder,
+		},
+		{
+			name: "ok - validation block weak parent ids unsorted",
```
muXxer: Seems like it....
```suggestion
			name: "fail - validation block weak parent ids unsorted",
```

output.go
```diff
@@ -700,8 +701,107 @@ func OutputsSyntacticalImplicitAccountCreationAddress() OutputsSyntacticalValida
...
+		elementValidationFunc := LexicalOrderAndUniquenessValidator[UnlockCondition]()
+		switch typedOutput := output.(type) {
+		case *BasicOutput:
+			for idx, uc := range typedOutput.UnlockConditions {
+				if err := elementValidationFunc(idx, uc); err != nil {
+					return err
+				}
+			}
+		case *FoundryOutput, *AccountOutput, *DelegationOutput:
+			// These outputs only have one address type.
```
muXxer: We should check anyway because of uniqueness.

muXxer (changes requested):

muXxer (approved):

PR #669, opened 2024-01-24
Align transaction builder mana calculation to vm

jkrvivian:


cyberphysic4l (commented):
I don't see how we can remove all these checks for the account-bound Mana. Could you explain why these should be removed?

muXxer:
The logic needs to remain as it is. Mana can be bound to accounts, and we need to take this into consideration.
See https://github.com/iotaledger/iota.go/blob/42c1479d3f301ce7ca44871edebd4283727503ac/vm/nova/vm.go#L382

PR #675, opened 2024-01-25
Fix block workscore and transaction builder utility funcs

### LINKED ISSUE #695: Fix mana calculation mismatch in transaction builder
The node returned mana amount mismatch on the native token creation transaction in `accounttransition_test.go` of docker testframework.

This should not happen after building a transaction with `AllotAllMana` option.

muXxer:


block.go
```diff
@@ -251,12 +251,10 @@ func (b *Block) WorkScore() (WorkScore, error) {
...
-	workScoreSignature, err := b.Signature.WorkScore(workScoreParameters)
-	if err != nil {
-		return 0, err
-	}
-
-	return workScoreBody.Add(workScoreSignature)
+	// the workscore of the block only consists of the workscore of the block body
+	// because the body should already include an offset for the "block",
+	// which accounts for the signature check of the block as well.
+	return workScoreBody, nil
```
PhilippGackstatter: How about we do

```go
func (b *Block) WorkScore() (WorkScore, error) {
	workScoreParameters := b.API.ProtocolParameters().WorkScoreParameters()
	// the workscore of the block only consists of the workscore of the block body
	// because the body should already include an offset for the "block",
	// which accounts for the signature check of the block as well.
	return b.Body.WorkScore(workScoreParameters)
}
```

meaning, we just pass the work score impl through to the body. Validation Block returns `0` and basic block body does what it does now.

We could even consider removing the `WorkScore` func on the top level block, but because we have access to the `API` here, it makes more sense to do it the way I suggested, I think.

PhilippGackstatter (approved):

PR #696, opened 2024-02-21
Retainer data cleanup and documentation

### LINKED ISSUE #723: Clearly define the meaning of the different flags (accepted, committed, finalized etc)
we should define that in the documentation and also check correctness in the code.

cyberphysic4l:
Rename, restructure and document the various block and transaction states to be stored in the retainer.

- BlockStateRejected renamed to align with our usual terminology, now called BlockStateOrphaned is returned when a the block's slot is committed, but the block is not yet accepted (it is therefore orphaned).
- Document meaning of the block and transaction states more clearly.

To be merged alongside https://github.com/iotaledger/iota-core/pull/766 
closes https://github.com/iotaledger/iota-core/issues/723


api/core.go
```diff
@@ -9,21 +9,25 @@ import (
...
 type BlockState byte
-type BlockFailureReason byte
-
-const (
-	BlockStateLength         = serializer.OneByte
-	BlockFailureReasonLength = serializer.OneByte
-)
 
 const (
+	// BlockStateUnknown indicates that the state of the block is can not be determined by the node for some reason.
```
PhilippGackstatter:
```suggestion
	// BlockStateUnknown indicates that the state of the block can not be determined by the node for some reason.
```

PhilippGackstatter (approved):

PR #697, opened 2024-02-23
Use same address key for implicit account address and ed25519 address from same pubkey

cyberphysic4l:
Use same address key for implicit account address and ed25519 address from same pubkey so both the address signer and the vm can recognise that a signature from the underlying pubkey is a valid unlock for either an implicit account address or an ed25519 address type.


muXxer:
That's not a solution IMO. The changed code is autogenerated, and I think we should rather change the logic inside the unlock. I will think about it a bit.

PR #698, opened 2024-02-23
Fix reference unlock by same public key but different address

muXxer:
My take on #697.

Currently it is impossible to unlock an implicit account creation address and an ed25519 address which are derived from the same public key in the same transaction. This is because we disallow to have the same public key in different signatures, but we also fail to check correctly for a valid reference unlock, if the underlying public key is the same, but the address is different.

This PR changes the logic in a way that the public key that unlocked the identity gets tracked, and in a reference unlock it is checked that the specific derived address for that public key matches.

The PR introduces the concept of a `SignerUID` in the `Signature` and `AddressSigner` interface to identify unique signatures signers (in out case public keys).

The `AddressSigner` for the `TransactionBuilder` now needs to be passed in the constructor instead of the `Build` method. This breaking change was necessary because otherwise the workscore for transaction payload was wrong in certain edge cases, because the formerly used `EmptyAddressSigner` had no knowledge about the `SignerUID`s and therefore didn't construct the `ReferentialUnlocks` properly. The `EmptyAddressSigner` was therefore removed.

The PR also removes the unused `Key` interface from the chain ID and other IDs, because the `Key` method of the underlying `Address` is used everywhere instead, which is more precise.

### DIFF
--- a/builder/transaction_builder.go
+++ b/builder/transaction_builder.go
@@ -10,9 +10,10 @@ import (
 var ErrTransactionBuilder = ierrors.New("transaction builder error")
 
 // NewTransactionBuilder creates a new TransactionBuilder.
-func NewTransactionBuilder(api iotago.API) *TransactionBuilder {
+func NewTransactionBuilder(api iotago.API, signer iotago.AddressSigner) *TransactionBuilder {
 	return &TransactionBuilder{
-		api: api,
+		api:    api,
+		signer: signer,
 		transaction: &iotago.Transaction{
 			API: api,
 			TransactionEssence: &iotago.TransactionEssence{
@@ -33,6 +34,7 @@ func NewTransactionBuilder(api iotago.API) *TransactionBuilder {
 // TransactionBuilder is used to easily build up a SignedTransaction.
 type TransactionBuilder struct {
 	api              iotago.API
+	signer           iotago.AddressSigner
 	occurredBuildErr error
 	transaction      *iotago.Transaction
 	inputs           iotago.OutputSet
@@ -58,10 +60,12 @@ func (b *TransactionBuilder) Clone() *TransactionBuilder {
 
 	return &TransactionBuilder{
 		api:              b.api,
+		signer:           b.signer,
 		occurredBuildErr: b.occurredBuildErr,
 		transaction:      b.transaction.Clone(),
 		inputs:           b.inputs.Clone(),
 		inputOwner:       cpyInputOwner,
+		rewards:          b.rewards,
 	}
 }
 
@@ -458,9 +462,9 @@ func (b *TransactionBuilder) hasManalockCondition(output iotago.Output) (iotago.
 
 // BuildAndSwapToBlockBuilder builds the transaction and then swaps to a BasicBlockBuilder with
 // the transaction set as its payload. txFunc can be nil.
-func (b *TransactionBuilder) BuildAndSwapToBlockBuilder(signer iotago.AddressSigner, txFunc TransactionFunc) *BasicBlockBuilder {
+func (b *TransactionBuilder) BuildAndSwapToBlockBuilder(txFunc TransactionFunc) *BasicBlockBuilder {
 	blockBuilder := NewBasicBlockBuilder(b.api)
-	tx, err := b.Build(signer)
+	tx, err := b.Build()
 	if err != nil {
 		blockBuilder.err = err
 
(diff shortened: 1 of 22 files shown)


jkrvivian (dismissed):

builder/transaction_builder.go
```diff
@@ -676,62 +690,127 @@ func (b *TransactionBuilder) Build(signer iotago.AddressSigner) (*iotago.SignedT
...
+			unlockedSet.addSignerUID(signerUID, inputIndex)
+		} else {
+			// add a referential unlock to the former unlock position
+			unlockedSet.addReferentialUnlock(owner, unlockedAtIndex)
 		}
 
-		unlocks = addReferentialUnlock(addr, unlocks, pos)
-		addChainAsUnlocked(inputs[i], i, unlockPos)
+		// always mark the chain as unlocked in case the output is a chain output
+		unlockedSet.addChainAsUnlocked(inputs[inputIndex], inputIndex)
```
cyberphysic4l, line 762: Can we have chain outputs unlocked by chain address? I recall something about this not being allowed...

vm/vm.go
```diff
@@ -155,50 +155,121 @@ func RunVMFuncs(vm VirtualMachine, vmParams *Params, execFuncs ...ExecFunc) erro
 	return nil
 }
 
+// unlockedAddressesSet holds a set of unlocked addresses.
+type unlockedAddressesSet struct {
+	// SignatureUnlockedAddrsByIndex contains direct unlockable addresses only which are unlocked by a signature,
+	// indexed by the index of the input/unlock index.
+	SignatureUnlockedAddrsByIndex map[uint16]*unlockedAddressWithSignature
+	// UnlockedAddrsByAddr contains all unlocked addresses.
+	UnlockedAddrsByAddr UnlockedAddresses
```
cyberphysic4l:
```suggestion
	UnlockedAddrsByAddrKey UnlockedAddresses
```

vm/vm.go
```diff
@@ -155,50 +155,121 @@ func RunVMFuncs(vm VirtualMachine, vmParams *Params, execFuncs ...ExecFunc) erro
...
+
+	return nil
+}
+
+// ReferentialUnlockDirectlyUnlockable expects a directly unlockable address and performs a check whether the given address
+// is unlocked at referenceInputIndex and if the signature of the referenced unlock matches the given address.
+// If all checks are successful, it adds the input index to the set of unlocked inputs by this address.
+// In case the given address is not yet unlocked, it is added to the set of unlocked addresses.
+// This is necessary if for example the signature was used before to unlock another address (e.g. derived from the same public key).
+func (s *unlockedAddressesSet) ReferentialUnlockDirectlyUnlockable(owner iotago.DirectUnlockableAddress, inputIndex uint16, referenceInputIndex uint16, checkUnlockOnly bool) error {
```
cyberphysic4l:
```suggestion
func (s *unlockedAddressesSet) ReferentialUnlockDirectlyUnlockable(owner iotago.DirectUnlockableAddress, inputIndex uint16, referencedInputIndex uint16, checkUnlockOnly bool) error {
```

signature.go
```diff
@@ -15,6 +15,13 @@ type Signature interface {
 	ProcessableObject
 	constraints.Cloneable[Signature]
 
+	// MatchesAddress returns true if the Signature matches the given Address.
+	MatchesAddress(Address) bool
+
+	// SignerUID returns the signer unique identifier for the signature.
```
cyberphysic4l:
```suggestion
	// SignerUID returns the unique identifier of the signature's signer.
```

signature_ed25519.go
```diff
@@ -60,6 +60,24 @@ func (e *Ed25519Signature) String() string {
...
+	case *Ed25519Address:
+		return targetAddress.Equal(Ed25519AddressFromPubKey(ed25519.PublicKey(e.PublicKey[:])))
+	case *ImplicitAccountCreationAddress:
+		return targetAddress.Equal(ImplicitAccountCreationAddressFromPubKey(ed25519.PublicKey(e.PublicKey[:])))
+	default:
+		return false
+	}
+}
+
+// SignerUID returns the signer unique identifier for the signature.
```
cyberphysic4l:
```suggestion
// SignerUID returns the unique identifier of the signature's signer.
```

unlock.go
```diff
@@ -156,11 +146,17 @@ type UnlockValidatorFunc func(index int, unlock Unlock) error
 //  6. MultiUnlock(s) are unique
 //  7. ReferenceUnlock(s) to MultiUnlock(s) are not nested in MultiUnlock(s)
 func SignatureUniqueAndReferenceUnlocksValidator(api API) UnlockValidatorFunc {
```
cyberphysic4l:
```suggestion
func SignaturesUniqueAndReferenceUnlocksValidator(api API) UnlockValidatorFunc {
```

cyberphysic4l (approved):

cyberphysic4l (approved):
Just a few small renames and comments suggested, but the logic changes seem quite minimal now that it's cleaned up so much and all looks good to me.

builder/transaction_builder.go
```diff
@@ -636,26 +640,36 @@ func (b *TransactionBuilder) MinRequiredAllottedMana(rmc iotago.Mana, blockIssue
...
 	return dummyBlock.ManaCost(rmc)
 }
 
 // Build signs the inputs with the given signer and returns the built payload.
-func (b *TransactionBuilder) Build(signer iotago.AddressSigner) (*iotago.SignedTransaction, error) {
+func (b *TransactionBuilder) Build() (*iotago.SignedTransaction, error) {
+	return b.build(true)
+}
+
+// Build signs the inputs with the given signer and returns the built payload.
```
alexsporn: adapt the comment

builder/transaction_builder.go
```diff
@@ -676,62 +690,131 @@ func (b *TransactionBuilder) Build(signer iotago.AddressSigner) (*iotago.SignedT
...
 	if chainInput, is := input.(iotago.ChainOutput); is && chainInput.ChainID().Addressable() {
-		prevUnlocked[chainInput.ChainID().ToAddress().Key()] = posUnlocked
+		u.unlockedChains[chainInput.ChainID().ToAddress().Key()] = unlockedAtIndex
 	}
 }
+
+// isChainUnlocked checks if the underlying chain address is unlocked and returns the unlock index.
+func (u *txBuilderUnlockedSet) isChainUnlocked(chainAddr iotago.ChainAddress) (int, bool) {
+	unlockedAtIndex, isUnlocked := u.unlockedChains[chainAddr.ChainID().ToAddress().Key()]
+	return unlockedAtIndex, isUnlocked
```
alexsporn, line 820: return directly, put the names into the return types to explain the signature

alexsporn (approved):
lgtm, makes it simpler to read 👍

builder/transaction_builder.go
muXxer: Unfortunately the go compiler doesn't understand that it needs to return two values if I return directly.
I will leave it as it is.

PR #709, opened 2024-03-08
Fix error message order and reconsider panics

PhilippGackstatter:
# Description of change

Fix error message order (`ierrors.Wrap` vs `ierrors.WithMessage`) and reconsider panicking in appropriate situations.

Implements part of iotaledger/iota-core#798.

## Notes on Error Handling

We have basically three methods of error handling available in Go:
- Named Errors
- Anonymous Errors
- Panics
### Named Errors

A _named error_ is one that is initialized as a static variable, e.g.:
```go
ErrTxEssenceNetworkIDInvalid = ierrors.New("invalid network ID")
```

It is part of the public interface and can therefore be matched against, using, for example, `ierrors.Is`.
This one should therefore be used whenever we want either users to be able to match against the error or we want to test the specific case in which the error occurs.

### Anonymous Errors

Because the idiomatic error handling in Go returns the `error` interface, the returned value can be opaque. All we know is that we can call `Error()` on the value and get a string representation for human consumption. An anonymous error is constructed ad-hoc, e.g. with `ierrors.New` or `ierrors.Errorf`:
```go
func VersionFromBytes(b []byte) (Version, int, error) {
	if len(b) < VersionLength {
		return 0, 0, ierrors.New("invalid version bytes length")
	}

	return Version(b[0]), VersionLength, nil
}
```
Anonymous errors should be used for cases where we do not want to panic but we do not need the caller to be able to match against the returned error. This is the case for broken invariants that _shouldn't_ happen, such as passing an incorrect length to `VersionFromBytes`. The error is intended to be read by a developer to figure out why their usage of the API was incorrect during development. It should not happen in a production deployment and therefore does not need to be matched against programmatically.

Deciding between named and anonymous errors is never easy. Anonymous errors make it harder to test against. The only way to make sure that an incorrect length passed to `VersionFromBytes` does return an error is to assert that _an_ error is returned or matching against the string, both of which are bad practice. On the other hand, making all errors named bloats the public interface.

### Panics

Panics are a lot like anonymous errors in that they do not have a public representation in the API and that they represent broken invariants. The major difference between returning an error and panicking is that the caller needs to deal with the returned error and that a panic takes away control from the caller.
Because a panics takes away control it must be used judiciously. A panic says: "this problem is serious enough that 'everything' should come to a halt."
Panics are okay to use when an essential invariant of a type or the system as a whole is broken. A concrete example is when we switch over all `Address` types (the Go types to be specific, not `iotago.AddressType`) and land in the `default` case. If we indeed handled all supported addresses then panicking in the `default` case is ok, because we can only run into that case if we added a new address type and have not handled it above or a user passed a type implementing the address interface (which is only possible when iota.go is used as a library). In both cases we want to panic to alert us to the unhandled type or the user to the unsupported type.

We must not panic in cases when users can construct input that would cause a panic, at least not when the input is passed to iota-core and then to iota.go, rather than to iota.go directly (e.g. as a library).
Because it is generally hard to verify that users can never do that for a given case, erring (no pun intended) on the side of caution and returning anonymous errors tends to be safer.
## Error Wrapping
We have two ways of wrapping errors, `Wrap` and `WithMessage` (and their corresponding formatting versions).

The usage is very simple.
`Wrap` for _prepending_ a message:
```go
return ierrors.Wrapf(iotago.ErrSenderFeatureNotUnlocked, "output %d", outputIndex)
```
Which results in the error message:
```
output 1: sender feature is not unlocked
```

`WithMessage` for _appending_ a message:
```go
return ierrors.WithMessagef(iotago.ErrStakingEndEpochTooEarly, "end epoch %d should be >= %d", stakingFeat.EndEpoch, unbondingEpoch)
```
Which results in the error message:
```
staking end epoch must be set to the transaction epoch plus the unbonding period: end epoch 29 should be >= 30
```

### Combining existing errors
Sometimes two existing errors need to be combined. Using `Join` is not desirable here, as it is, in my understanding, intended to group together a collection of errors. However, in our use cases we almost always want to wrap one error in another to create a hierarchical chain of errors. One example of this is:

```go
totalIn, err = safemath.SafeAdd(totalIn, manaStored)
if err != nil {
    return 0, ierrors.WithMessagef(iotago.ErrManaOverflow, "%w", err)
}
```

We want to return `ErrManaOverflow` as a higher-level error for all kinds of overflows that can happen with Mana. This error is later mapped to its own transaction failure reason, `TxFailureManaOverflow`. We also want to include the underlying safemath error in the error chain so we get the detailed error of which calculation went wrong:
```
under- or overflow in Mana calculations: integer under- or overflow: 18446744073709551615 + 10
```

If we were to use `Join` we would not create a chain of errors but only group the errors together, which is reflected in the error message as a newline, separating the two errors:

```
under- or overflow in Mana calculations
integer under- or overflow: 18446744073709551615 + 10
```

This makes it look like they are unrelated and it is only clear from context here that they are in fact related.

We might want to add a dedicated function in `ierrors` for this, perhaps `ierrors.WithError(iotago.ErrManaOverflow, err)`, analogous to `WithMessage`.


vm/vm.go
```diff
@@ -78,19 +77,19 @@ func TotalManaIn(manaDecayProvider *iotago.ManaDecayProvider, storageScoreStruct
 		// stored Mana
 		manaStored, err := manaDecayProvider.DecayManaBySlots(input.StoredMana(), outputID.CreationSlot(), txCreationSlot)
 		if err != nil {
-			return 0, ierrors.Wrapf(err, "input %s stored mana calculation failed", outputID)
+			return 0, ierrors.Wrapf(err, "stored mana calculation failed for input %s", outputID)
 		}
 		totalIn, err = safemath.SafeAdd(totalIn, manaStored)
 		if err != nil {
-			return 0, ierrors.Wrapf(iotago.ErrManaOverflow, "%w", err)
+			return 0, ierrors.WithMessagef(iotago.ErrManaOverflow, "%w", err)
```
muXxer, line 84: why not `ierrors.Join()` here?

muXxer: Please also change that for all the following occurrences in this file.

vm/vm.go
```diff
@@ -437,12 +436,12 @@ func ValidateUnlocks(signedTransaction *iotago.SignedTransaction, resolvedInputs
 
 	txID, err := signedTransaction.Transaction.ID()
 	if err != nil {
-		return nil, ierrors.Wrapf(err, "failed to compute transaction ID")
+		panic(fmt.Sprintf("transaction ID computation should have succeeded: %s", err))
```
muXxer: does %s print the correct string for an Error? I'm not sure every error has a "String" method implemented. You should maybe call `err.Error()` to be sure?

vm/vm.go
```diff
@@ -437,12 +436,12 @@ func ValidateUnlocks(signedTransaction *iotago.SignedTransaction, resolvedInputs
...
 	}
 
 	essenceMsgToSign, err := signedTransaction.Transaction.SigningMessage()
 	if err != nil {
-		return nil, ierrors.Wrapf(err, "failed to compute signing message")
+		panic(fmt.Sprintf("signing message computation should have succeeded: %s", err))
```
muXxer: same here

vm/nova/vm.go
```diff
@@ -42,19 +42,14 @@ func NewVMParamsWorkingSet(api iotago.API, t *iotago.Transaction, resolvedInputs
...
 
-	workingSet.EssenceMsgToSign, err = t.SigningMessage()
-	if err != nil {
-		return nil, err
-	}
-
 	txID, err := workingSet.Tx.ID()
 	if err != nil {
-		return nil, err
+		panic(fmt.Sprintf("transaction ID computation should have succeeded: %s", err))
```
muXxer:
```suggestion
		panic(fmt.Sprintf("transaction ID computation should have succeeded: %s", err.Error()))
```

address.go
```diff
@@ -175,24 +173,24 @@ type UTXOIDChainID interface {
...
+		return &ImplicitAccountCreationAddress{}
 	case AddressMulti:
-		return &MultiAddress{}, nil
+		return &MultiAddress{}
 	case AddressRestricted:
-		return &RestrictedAddress{}, nil
+		return &RestrictedAddress{}
 	default:
-		return nil, ierrors.Wrapf(ErrUnknownAddrType, "type %d", addressType)
+		panic(fmt.Sprintf("unknown address type %d", addressType))
```
muXxer: With that change, the `ParseBech32` which takes user input could panic on unknown address types.
I don't think this is a good idea.

address.go
```diff
@@ -312,6 +307,6 @@ func AddressFromReader(reader io.ReadSeeker) (Address, error) {
 		return RestrictedAddressFromReader(reader)
 
 	default:
-		return nil, ierrors.Wrapf(ErrUnknownAddrType, "type %d", addressType)
+		panic(fmt.Sprintf("unknown address type %d", addressType))
```
muXxer: I also wouldn't panic here.

vm/vm.go
```diff
@@ -753,16 +749,16 @@ func ExecFuncBalancedBaseTokens() ExecFunc {
...
 		if in != out {
-			return ierrors.Wrapf(iotago.ErrInputOutputBaseTokenMismatch, "in %d, out %d", in, out)
+			return ierrors.WithMessagef(iotago.ErrInputOutputBaseTokenMismatch, "in %d, out %d", in, out)
 		}
 
 		for addr, returnSum := range inputSumReturnAmountPerAddress {
 			outSum, has := outputSimpleTransfersPerAddr[addr]
 			if !has {
-				return ierrors.Wrapf(iotago.ErrReturnAmountNotFulFilled, "return amount of %d not fulfilled as there is no output for %s", returnSum, addr)
+				return ierrors.WithMessagef(iotago.ErrReturnAmountNotFulFilled, "return amount of %d not fulfilled as there is no output for address (serialized) %s", returnSum, hexutil.EncodeHex([]byte(addr)))
```
muXxer, line 758: Why not use `addr.Bech32(vmParams.API.ProtocolParameters().Bech32HRP())`

vm/vm.go
```diff
@@ -753,16 +749,16 @@ func ExecFuncBalancedBaseTokens() ExecFunc {
...
 			}
 			if outSum < returnSum {
-				return ierrors.Wrapf(iotago.ErrReturnAmountNotFulFilled, "return amount of %d not fulfilled as output is only %d for %s", returnSum, outSum, addr)
+				return ierrors.WithMessagef(iotago.ErrReturnAmountNotFulFilled, "return amount of %d not fulfilled as output is only %d for address (serialized) %s", returnSum, outSum, hexutil.EncodeHex([]byte(addr)))
```
muXxer, line 761: here as well

vm/vm.go
```diff
@@ -798,13 +794,13 @@ func ExecFuncChainTransitions() ExecFunc {
 			next := vmParams.WorkingSet.OutChains[chainID]
 			if next == nil {
 				if err := vm.ChainSTVF(vmParams, iotago.ChainTransitionTypeDestroy, inputChain, nil); err != nil {
-					return ierrors.Join(iotago.ErrChainTransitionInvalid, ierrors.Wrapf(err, "input chain %s (%T) destruction transition failed", chainID, inputChain))
+					return ierrors.Wrapf(err, "invalid destruction for %s %s", inputChain.Output.Type(), chainID)
```
muXxer, line 797:
```suggestion
					return ierrors.Wrapf(err, "invalid destruction for %s %s", inputChain.Output.Type(), chainID.ToHex())
```

vm/vm.go
```diff
@@ -798,13 +794,13 @@ func ExecFuncChainTransitions() ExecFunc {
...
 				}
 
 				continue
 			}
 			if err := vm.ChainSTVF(vmParams, iotago.ChainTransitionTypeStateChange, inputChain, next); err != nil {
-				return ierrors.Join(iotago.ErrChainTransitionInvalid, ierrors.Wrapf(err, "chain %s (%T) state transition failed", chainID, inputChain))
+				return ierrors.Wrapf(err, "invalid transition for %s %s", inputChain.Output.Type(), chainID)
```
muXxer, line 803:
```suggestion
				return ierrors.Wrapf(err, "invalid transition for %s %s", inputChain.Output.Type(), chainID.ToHex())
```

vm/vm.go
```diff
@@ -814,7 +810,7 @@ func ExecFuncChainTransitions() ExecFunc {
 			}
 
 			if err := vm.ChainSTVF(vmParams, iotago.ChainTransitionTypeGenesis, nil, outputChain); err != nil {
-				return ierrors.Join(iotago.ErrChainTransitionInvalid, ierrors.Wrapf(err, "new chain %s (%T) state transition failed", chainID, outputChain))
+				return ierrors.Wrapf(err, "invalid creation of %s %s", outputChain.Type(), chainID)
```
muXxer, line 813:
```suggestion
				return ierrors.Wrapf(err, "invalid creation of %s %s", outputChain.Type(), chainID.ToHex())
```

vm/vm.go
```diff
@@ -831,12 +827,12 @@ func ExecFuncBalancedNativeTokens() ExecFunc {
...
 		vmParams.WorkingSet.InNativeTokens, err = vmParams.WorkingSet.UTXOInputs.NativeTokenSum()
 		if err != nil {
-			return ierrors.Join(iotago.ErrNativeTokenSetInvalid, ierrors.Errorf("invalid input native token set: %w", err))
+			return ierrors.WithMessagef(iotago.ErrNativeTokenSetInvalid, "invalid input native token set: %w", err)
 		}
 
 		vmParams.WorkingSet.OutNativeTokens, err = vmParams.WorkingSet.Tx.Outputs.NativeTokenSum()
 		if err != nil {
-			return ierrors.Join(iotago.ErrNativeTokenSetInvalid, err)
+			return ierrors.WithMessagef(iotago.ErrNativeTokenSetInvalid, "%w", err)
```
muXxer, line 835: Also in this file, better use ierrors.Join

vm/vm.go
```diff
@@ -850,10 +846,10 @@ func ExecFuncBalancedNativeTokens() ExecFunc {
 
 			if vmParams.WorkingSet.Tx.Capabilities.CannotBurnNativeTokens() && (outSum == nil || inSum.Cmp(outSum) != 0) {
 				// if burning is not allowed, the input sum must be equal to the output sum
-				return ierrors.Wrapf(iotago.ErrTxCapabilitiesNativeTokenBurningNotAllowed, "%w: native token %s is less on output (%d) than input (%d) side but burning is not allowed in the transaction and the foundry is absent for melting", iotago.ErrNativeTokenSumUnbalanced, nativeTokenID, outSum, inSum)
+				return ierrors.WithMessagef(iotago.ErrTxCapabilitiesNativeTokenBurningNotAllowed, "%w: native token %s is less on output (%d) than input (%d) side but burning is not allowed in the transaction and the foundry is absent for melting", iotago.ErrNativeTokenSumUnbalanced, nativeTokenID, outSum, inSum)
```
muXxer, line 849: please add `.ToHex()` to all the `nativeTokenID`

vm/nova/vm.go
```diff
@@ -42,19 +42,14 @@ func NewVMParamsWorkingSet(api iotago.API, t *iotago.Transaction, resolvedInputs
 		workingSet.InputIDToInputIndex[txInputID] = uint16(inputIndex)
 		input, ok := workingSet.UTXOInputsSet[txInputID]
 		if !ok {
-			return nil, ierrors.Wrapf(iotago.ErrMissingUTXO, "utxo for input %d not supplied", inputIndex)
+			panic(fmt.Sprintf("UTXO for input %d should be supplied %s", inputIndex, txInputID))
```
muXxer, line 45:
```suggestion
			panic(fmt.Sprintf("UTXO for input %d should be supplied %s", inputIndex, txInputID.ToHex()))
```

vm/nova/vm.go
```diff
@@ -886,19 +861,18 @@ func foundrySerialNumberValid(vmParams *vm.Params, current *iotago.FoundryOutput
...
+		)
 	}
 
 	// the check for the serial number and token scheme not being mutated is implicit
 	// as a change would cause the foundry ID to be different, which would result in
 	// no matching foundry to be found to validate the state transition against
 	if current.MustFoundryID() != next.MustFoundryID() {
 		// impossible invariant as the STVF should be called via the matching next foundry output
-		panic(fmt.Sprintf("foundry IDs mismatch in state transition validation function: have %v got %v", current.MustFoundryID(), next.MustFoundryID()))
+		panic(fmt.Sprintf("foundry IDs mismatch in state transition validation function: have %s got %s", current.MustFoundryID(), next.MustFoundryID()))
```
muXxer:
```suggestion
		panic(fmt.Sprintf("foundry IDs mismatch in state transition validation function: have %s got %s", current.MustFoundryID().ToHex(), next.MustFoundryID().ToHex()))
```

vm/nova/vm.go
```diff
@@ -873,11 +849,10 @@ func foundrySerialNumberValid(vmParams *vm.Params, current *iotago.FoundryOutput
...
 
 		if otherFoundryOutput.SerialNumber >= current.SerialNumber {
-			return ierrors.Join(iotago.ErrInvalidFoundryStateTransition,
-				ierrors.WithMessagef(
-					iotago.ErrFoundrySerialInvalid,
-					"new foundry output %s at index %d has bigger equal serial number than this foundry %s", otherFoundryID, outputIndex, thisFoundryID,
-				))
+			return ierrors.WithMessagef(
+				iotago.ErrFoundrySerialInvalid,
+				"new foundry output %s at index %d has bigger equal serial number than this foundry %s", otherFoundryID, outputIndex, thisFoundryID,
```
muXxer:
```suggestion
				"new foundry output %s at index %d has bigger equal serial number than this foundry %s", otherFoundryID.ToHex(), outputIndex, thisFoundryID.ToHex(),
```

vm/nova/vm.go
```diff
@@ -839,11 +816,10 @@ func foundrySerialNumberValid(vmParams *vm.Params, current *iotago.FoundryOutput
...
 	endIncSerial := outAccount.FoundryCounter
 	if startSerial >= current.SerialNumber || current.SerialNumber > endIncSerial {
-		return ierrors.Join(iotago.ErrInvalidFoundryStateTransition,
-			ierrors.WithMessagef(
-				iotago.ErrFoundrySerialInvalid,
-				"new foundry output %s's serial number %d is not between the foundry counter interval of [%d,%d)", thisFoundryID, current.SerialNumber, startSerial, endIncSerial,
-			))
+		return ierrors.WithMessagef(
+			iotago.ErrFoundrySerialInvalid,
+			"new foundry output %s's serial number %d is not between the foundry counter interval of [%d,%d)", thisFoundryID, current.SerialNumber, startSerial, endIncSerial,
```
muXxer:
```suggestion
			"new foundry output %s's serial number %d is not between the foundry counter interval of [%d,%d)", thisFoundryID.ToHex(), current.SerialNumber, startSerial, endIncSerial,
```

vm/nova/vm.go
```diff
@@ -816,18 +795,16 @@ func foundryGenesisValid(vmParams *vm.Params, current *iotago.FoundryOutput, thi
...
 	}
 
 	outAccount, ok := vmParams.WorkingSet.OutChains[accountID]
 	if !ok {
-		return ierrors.Join(iotago.ErrInvalidFoundryStateTransition,
-			ierrors.WithMessagef(iotago.ErrFoundryTransitionWithoutAccount,
-				"missing output transitioning account output %s for new foundry output %s", accountID, thisFoundryID,
-			))
+		return ierrors.WithMessagef(iotago.ErrFoundryTransitionWithoutAccount,
+			"missing output transitioning account output %s for new foundry output %s", accountID, thisFoundryID,
```
muXxer:
```suggestion
			"missing output transitioning account output %s for new foundry output %s", accountID.ToHex(), thisFoundryID.ToHex(),
```

vm/nova/vm.go
```diff
@@ -816,18 +795,16 @@ func foundryGenesisValid(vmParams *vm.Params, current *iotago.FoundryOutput, thi
 	accountID := current.Owner().(*iotago.AccountAddress).AccountID()
 	inAccount, ok := vmParams.WorkingSet.InChains[accountID]
 	if !ok {
-		return ierrors.Join(iotago.ErrInvalidFoundryStateTransition,
-			ierrors.WithMessagef(iotago.ErrFoundryTransitionWithoutAccount,
-				"missing input transitioning account output %s for new foundry output %s", accountID, thisFoundryID,
-			))
+		return ierrors.WithMessagef(iotago.ErrFoundryTransitionWithoutAccount,
+			"missing input transitioning account output %s for new foundry output %s", accountID, thisFoundryID,
```
muXxer:
```suggestion
			"missing input transitioning account output %s for new foundry output %s", accountID.ToHex(), thisFoundryID.ToHex(),
```

vm/nova/vm.go
```diff
@@ -755,15 +734,15 @@ func anchorStateSTVF(input *vm.ChainOutputWithIDs, next *iotago.AnchorOutput) er
 	current := input.Output.(*iotago.AnchorOutput)
 	switch {
 	case !current.StateController().Equal(next.StateController()):
-		return ierrors.Wrapf(iotago.ErrAnchorInvalidStateTransition, "state controller changed, in %v / out %v", current.StateController(), next.StateController())
+		return ierrors.WithMessagef(iotago.ErrAnchorInvalidStateTransition, "state controller changed, in %v / out %v", current.StateController(), next.StateController())
```
muXxer: can you use the bech32 of all the addresses? the vmParams contain the protocol params.

input.go
```diff
@@ -90,11 +90,11 @@ func InputsSyntacticalUnique() ElementValidationFunc[Input] {
 			referencedOutputID := castInput.OutputID()
 			k := string(referencedOutputID[:])
 			if j, has := utxoSet[k]; has {
-				return ierrors.Wrapf(ErrInputUTXORefsNotUnique, "input %d and %d share the same referenced UTXO index", j, index)
+				return ierrors.WithMessagef(ErrInputUTXORefsNotUnique, "input %d and %d share the same referenced UTXO index", j, index)
 			}
 			utxoSet[k] = index
 		default:
-			return ierrors.Wrapf(ErrUnknownInputType, "input %d, tx can only contain UTXO inputs", index)
+			panic("all supported input types should be handled above")
```
muXxer, line 101: Can you add the same nice explanation with the go type as with the addresses?

input.go
```diff
@@ -108,10 +108,10 @@ func InputsSyntacticalIndicesWithinBounds() ElementValidationFunc[Input] {
 		case *UTXOInput:
 			// TODO: do we really want to check the max value on the input side?
 			if castInput.Index() < RefUTXOIndexMin || castInput.Index() > RefUTXOIndexMax {
-				return ierrors.Wrapf(ErrRefUTXOIndexInvalid, "input %d", index)
+				return ierrors.WithMessagef(ErrRefUTXOIndexInvalid, "input %d", index)
 			}
 		default:
-			return ierrors.Wrapf(ErrUnknownInputType, "input %d, tx can only contain UTXInput inputs", index)
+			panic("all supported input types should be handled above")
```
muXxer, line 122: here as well.

input_context.go
```diff
@@ -102,11 +97,11 @@ func ContextInputsRewardInputMaxIndex(inputsCount uint16) ElementValidationFunc[
 		case *RewardInput:
 			utxoIndex := castInput.Index
 			if utxoIndex >= inputsCount {
-				return ierrors.Wrapf(ErrInputRewardIndexExceedsMaxInputsCount, "reward input %d references index %d which is equal or greater than the inputs count %d",
+				return ierrors.WithMessagef(ErrInputRewardIndexExceedsMaxInputsCount, "reward input %d references index %d which is equal or greater than the inputs count %d",
 					index, utxoIndex, inputsCount)
 			}
 		default:
-			return ierrors.Wrapf(ErrUnknownContextInputType, "context input %d, tx can only contain CommitmentInputs, BlockIssuanceCreditInputs or RewardInputs", index)
+			panic("all supported context input types should be handled above")
```
muXxer, line 108: Can you add the same nice explanation with the go type as with the addresses?

input_context.go
```diff
@@ -126,14 +121,14 @@ func ContextInputsCommitmentInputRequirement() ElementValidationFunc[ContextInpu
...
+				return ierrors.WithMessagef(ErrCommitmentInputMissing, "block issuance credit input at index %d requires a commitment input", index)
 			}
 		case *RewardInput:
 			if !seenCommitmentInput {
-				return ierrors.Wrapf(ErrCommitmentInputMissing, "reward input at index %d requires a commitment input", index)
+				return ierrors.WithMessagef(ErrCommitmentInputMissing, "reward input at index %d requires a commitment input", index)
 			}
 		default:
-			return ierrors.Wrapf(ErrUnknownContextInputType, "context input %d, tx can only contain CommitmentInputs, BlockIssuanceCreditInputs or RewardInputs", index)
+			panic("all supported context input types should be handled above")
```
muXxer, line 139: Can you add the same nice explanation with the go type as with the addresses?

output.go
```diff
@@ -694,7 +668,7 @@ func OutputsSyntacticalImplicitAccountCreationAddress() ElementValidationFunc[Ou
 				return ierrors.WithMessagef(ErrImplicitAccountCreationAddressInInvalidOutput, "output %d", index)
 			}
 		default:
-			panic("unrecognized output type")
+			panic("all supported output types should be handled above")
```
muXxer, line 675: Can you add the same nice explanation with the go type as with the addresses?

output.go
```diff
@@ -743,7 +717,7 @@ func OutputsSyntacticalUnlockConditionLexicalOrderAndUniqueness() ElementValidat
 				}
 			}
 		default:
-			panic("unrecognized output type")
+			panic("all supported output types should be handled above")
```
muXxer, line 728: Can you add the same nice explanation with the go type as with the addresses?

output.go
```diff
@@ -812,7 +786,7 @@ func OutputsSyntacticalFeaturesLexicalOrderAndUniqueness() ElementValidationFunc
 			// This output does not have features.
 			return nil
 		default:
-			panic("unrecognized output type")
+			panic("all supported output types should be handled above")
```
muXxer, line 801: Can you add the same nice explanation with the go type as with the addresses?

testlog.txt
muXxer, line 1: What is this file?

transaction.go
```diff
@@ -133,18 +133,18 @@ func (t *Transaction) Clone() *Transaction {
...
-func (t *Transaction) Inputs() ([]*UTXOInput, error) {
+func (t *Transaction) Inputs() []*UTXOInput {
 	references := make([]*UTXOInput, 0, len(t.TransactionEssence.Inputs))
 	for _, input := range t.TransactionEssence.Inputs {
 		switch castInput := input.(type) {
 		case *UTXOInput:
 			references = append(references, castInput)
 		default:
-			return nil, ErrUnknownInputType
+			panic("all supported input types should be handled above")
```
muXxer, line 147: Can you add the same nice explanation with the go type as with the addresses?

transaction.go
```diff
@@ -161,21 +161,21 @@ func (t *Transaction) OutputsSet() (OutputSet, error) {
...
-func (t *Transaction) ContextInputs() (TransactionContextInputs, error) {
+func (t *Transaction) ContextInputs() TransactionContextInputs {
 	references := make(TransactionContextInputs, 0, len(t.TransactionEssence.ContextInputs))
 	for _, input := range t.TransactionEssence.ContextInputs {
 		switch castInput := input.(type) {
 		case *CommitmentInput, *BlockIssuanceCreditInput, *RewardInput:
 			references = append(references, castInput)
 		default:
-			return nil, ErrUnknownContextInputType
+			panic("all supported context input types should be handled above")
```
muXxer, line 179: Can you add the same nice explanation with the go type as with the addresses?

transaction.go
```diff
@@ -184,14 +184,14 @@ func (t *Transaction) BICInputs() ([]*BlockIssuanceCreditInput, error) {
 		case *CommitmentInput, *RewardInput:
 			// ignore this type
 		default:
-			return nil, ErrUnknownContextInputType
+			panic("all supported context input types should be handled above")
```
muXxer, line 199: Can you add the same nice explanation with the go type as with the addresses?

transaction.go
```diff
@@ -200,11 +200,11 @@ func (t *Transaction) RewardInputs() ([]*RewardInput, error) {
 		case *CommitmentInput, *BlockIssuanceCreditInput:
 			// ignore this type
 		default:
-			return nil, ErrUnknownContextInputType
+			panic("all supported context input types should be handled above")
```
muXxer, line 219: Can you add the same nice explanation with the go type as with the addresses?

transaction.go
```diff
@@ -216,7 +216,7 @@ func (t *Transaction) CommitmentInput() *CommitmentInput {
 		case *CommitmentInput:
 			return castInput
 		default:
-			return nil
+			panic("all supported context input types should be handled above")
```
muXxer, line 239: Can you add the same nice explanation with the go type as with the addresses?

unlock.go
```diff
@@ -240,29 +238,29 @@ func SignaturesUniqueAndReferenceUnlocksValidator(api API) UnlockValidatorFunc {
...
-					return ierrors.Wrapf(ErrNestedMultiUnlock, "unlock at index %d.%d is invalid", index, subIndex)
+					return ierrors.WithMessagef(ErrNestedMultiUnlock, "unlock at index %d.%d is invalid", index, subIndex)
 
 				case *EmptyUnlock:
 					// empty unlocks are allowed inside of multi unlocks
 					continue
 
 				default:
-					return ierrors.Wrapf(ErrUnknownUnlockType, "unlock at index %d.%d is of unknown type %T", index, subIndex, subUnlock)
+					panic("all supported unlock types should be handled above")
```
muXxer, line 257: Can you add the same nice explanation with the go type as with the addresses?

unlock.go
```diff
@@ -240,29 +238,29 @@ func SignaturesUniqueAndReferenceUnlocksValidator(api API) UnlockValidatorFunc {
...
 			seenMultiUnlocks[uint16(index)] = struct{}{}
 			seenMultiUnlockBytes[string(multiUnlockBytes)] = index
 
 		case *EmptyUnlock:
-			return ierrors.Wrapf(ErrEmptyUnlockOutsideMultiUnlock, "unlock at index %d is invalid", index)
+			return ierrors.WithMessagef(ErrEmptyUnlockOutsideMultiUnlock, "unlock at index %d is invalid", index)
 
 		default:
-			return ierrors.Wrapf(ErrUnknownUnlockType, "unlock at index %d is of unknown type %T", index, unlock)
+			panic("all supported unlock types should be handled above")
```
muXxer, line 271: Can you add the same nice explanation with the go type as with the addresses?

muXxer (changes requested):

vm/vm.go
PhilippGackstatter: `addr` is of type `string` because its the result of calling `Key()` on the return address. We could reconstruct the address but that seemed like a bit too much so I thought why not just print the serialized address (which is equivalent to `Key()`).

input.go
PhilippGackstatter: Btw just above this line I think you added a TODO: 
> // TODO: do we really want to check the max value on the input side?

I'm not fully sure what you mean by it, but if you mean why we check that the index is within bounds on the input side even though we only create indicies within that range on the output side: the index on the `UTXOInput` can be set by a user so it could be set to whatever value, so we should check here. Although not checking wouldn't be problematic either, because we just would run into a "input not found" error later.

vm/vm.go
PhilippGackstatter: Please see the PR description for why I chose to use this style over `Join`.

input.go
muXxer: We discussed that in the past, and we want to actually have that property, that we can inject bigger indexes in the snapshot, but the user should not be able to create it in a transaction. So we check it on the output side, but not on the input side.
As you said, we simply would run into "input not found" if a user crafts a transaction with a higher index that doesn't exist.

vm/vm.go
muXxer: ah ok, then fine :+1:

muXxer (approved):

